PR #90, opened 2025-01-09
feat: Implement `createContext` and `useContext`

### LINKED ISSUE #89: Use the `useContext` pattern to support strong typing for Providers.
Let's learn from React, and implement something similar to React.Context and the `useContext` hook to provider between typing and access for context.

jmoseley:
## Proposed changes

Implement a Context pattern reminiscent of the React pattern, leveraging a `createContext` and `useContext`. This improves type safety for contexts, and simplifies the references, making it easy to access the context from inside a component.

Fixes #89


packages/gensx/src/component.ts
```diff
@@ -66,12 +64,11 @@ export function Component<P, O>(
       const result = await resolveDeep(fn(props));
 
       let finalResult: O;
-      if (props.children) {
-        finalResult = await withContext({}, () =>
-          props.children?.(result as O),
-        );
+      if (props.children && !props.children.__gsxChildExecuted) {
+        props.children.__gsxChildExecuted = true;
```
jmoseley, line 68: I'm not super happy with the necessity to track whether children have been executed or not, but it was the only way to solve this at the moment.

The big issue is that (unlike React) we have "implicit" children, where the component wraps some children, but is not aware of them. We don't require a component to explicitly execute its children, it is simply handled by the framework.

However, there are moments (such as Fragments or Context providers) where the component does need to be aware of its children, and might choose how they are executed. Therefore, we need a way to know which is which, and this was the simplest way.

I expect a cleaner representation of this to become apparent in the near future (maybe via some of the checkpointing work), but it may be the case that this sort of tracking is necessary.

I also feel like there are a bunch of edge cases that this solution won't handle, but I wrote tests for all the interesting cases that I could think of, and they all passed :shrug:

EvanBoyle (commented):
@jmoseley as I understand it the issue is that the provider needs to hook in and adjust the way that it's children execute - so it manually executes it's children with a little additional context. 

I wonder if we could turn `ProviderComponent` into a first framework concept?

Is there a way that we could create a ProviderComponent primitive that had hooks into execute? 

If a component is a provider component, then before executing children, we would execute a predefined "before" callback or "provide" callback provided by the provider author. Then the framework could be aware if it was executing children within a provider and walk back up the "provide" stack to set appropriate context before calling children. 

Would this make sense rather than having to track this child execution? 

Maybe something like: 

```tsx
interface ProviderProps {
  children: JSX.Element | JSX.Element[];  // Direct children, not a function
}

function Provider<T>(
  hooks: {
    provide: () => T | Promise<T>;
    cleanup?: () => void | Promise<void>;
  }
) {
  return function ProviderComponent({ children }: ProviderProps) {
    return async () => {
      const context = await hooks.provide();
      
      try {
        return await withContext({ /* context */ }, 
          () => resolveDeep(children)  // Direct resolution of children
        );
      } finally {
        await hooks.cleanup?.();
      }
    };
  };
}
```

Good things about this IMO: 
- Clear separation between Providers and value-producing Components
- No execution tracking needed
- No ambiguity about child execution

What do you think?

jmoseley:
There still needs to be some magic inside `execute` or `resolveDeep` to ensure that children are not executed twice. And the issue is not with the Provider itself, but any wrapper around the provider (which are just plain Components). So we would actually need a `ProviderWrapper` that the framework treats specially, and that feels wonky

EvanBoyle:
What about something like this? 

```tsx
interface ProviderComponentConfig<P, C> {
  // Hook to transform/wrap children before execution
  transformChildren: (props: P, children: JSX.Element) => Promise<JSX.Element>;
  // Optional setup/cleanup hooks
  setup?: (props: P) => Promise<C>;
  cleanup?: () => Promise<void>;
}

const OpenAIProvider = gsx.ProviderComponent<ClientOptions, OpenAI>({
  setup: async (props) => new OpenAI(props),
  transformChildren: async (props, children) => (
    <OpenAIContext.Provider value={{ client: await setup(props) }}>
      {children}
    </OpenAIContext.Provider>
  )
});
```

I think the problem right now is that the `OpenAIProvider` that we export is itself a `gsx.Component` so of course the double execution happens. 

If we create a new concept `gsx.ProviderComponent` - then provider component could just have special casing. It doesn't get executed or produce outputs, it just transforms its children.

jmoseley:
I'm really not keen on the `ProviderComponent` idea, I think we are asking the developer to do too much, and pigeon holing the framework into supporting very narrow things. There may be other cases where someone wants to do this for some a reason besides `Providers` (wrap in a logger, error boundary, etc) and we don't want to have to special case those as well.

We could have a `ComponentWithChildren` to make this more general, but it does require the developer to use `props.children` or the rest of the tree won't be executed, and that would be a confusing failure mode.

I think it would be much better for the framework to do this transparently, so things just work with out the developer needing to understand these nuances.

EvanBoyle (commented):
🚢 Looks good! 

Do we also need to update our examples? HN, blog writer, streaming, etc?

jmoseley:
I will double check.

jmoseley:
The examples already used `OpenAIProvider`, so no updates necessary.

PR #110, opened 2025-01-14
feat: adding deep research example

dereklegenzoff:
Adds an example showing a "deep research" pattern. 

The workflow starts with a prompt and then goes through the following steps to generate a research report:

1. Converts the prompt into a series of queries
2. Uses the [Arxiv API](https://arxiv.org/help/api/user-manual) to retrieve relevant research papers
3. Determines if the papers are relevant to the prompt
4. Scrapes the full text of the relevant papers using the [Firecrawl API](https://www.firecrawl.dev/)
5. Summarizes the relevant parts of the papers
6. Uses the summaries to generate a research report


.github/workflows/test-examples.yml
```diff
@@ -44,3 +44,6 @@ jobs:
         run: pnpm build:examples
       - name: Run tests ✅
         run: pnpm test:examples
```
jmoseley, line 46: I think you should change the trigger on this (to `pull_request`) as well, so that we don't end up with this snafu next time.

examples/deepResearch/README.md
```diff
@@ -0,0 +1,35 @@
...
+- [OpenAI API Key](https://platform.openai.com/)
+
+## Usage
+
+```bash
+# Install dependencies
+pnpm install
+
+# Run the example
+FIRECRAWL_API_KEY=your_api_key OPENAI_API_KEY=your_api_key npm run run
```
jmoseley, line 34:
```suggestion
FIRECRAWL_API_KEY=your_api_key OPENAI_API_KEY=your_api_key pnpm run run
```

examples/deepResearch/deepResearch.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { PromptToQuery } from "./promptToQuery.js";
+import { FetchAndSummarize, FetchAndSummarizeOutput } from "./summarize.js";
+
+interface FindResearchProps {
+  prompt: string;
+}
+
+export const FindResearch = gsx.Component<FindResearchProps, ArxivEntry[]>(
+  ({ prompt }) => (
+    <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
```
jmoseley, line 15: You don't need to have this inside the component, you only need it in one place in the component tree (probably at the outermost point).

examples/deepResearch/summarize.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return {
+    summaries: documents.map((document) => {
+      const url = document.url.replace("abs", "html");
+      return {
+        title: document.title,
+        url: url,
+        summary: (
+          <Firecrawl url={url}>
+            {(markdown: string | null) =>
+              markdown && <SummarizePaper markdown={markdown} prompt={prompt} />
```
jmoseley, line 71: I love that this works

jmoseley (approved):
This looks great. Nice work. Sample output?

examples/deepResearch/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "gensx": "workspace:*",
+    "node-fetch": "^3.3.2",
+    "openai": "^4.77.0",
+    "xml2js": "^0.6.2"
+  },
+  "devDependencies": {
+    "@types/node": "^20.17.11",
+    "@types/xml2js": "^0.4.14",
+    "nodemon": "^3.1.9",
+    "prettier": "^3.0.0",
```
jmoseley, line 30: I notice it is in the other examples as well, but I don't think this is necessary, as we have it in the root package.json.

examples/deepResearch/deepResearch.tsx
dereklegenzoff: good catch! ended up removing this component completely

examples/deepResearch/package.json
dereklegenzoff: ahh makes sense. I've just been copying example folders and updating them so that explains the proliferation. I'll remove it in this one and go clean up the others at some point.

.github/workflows/test-examples.yml
dereklegenzoff: fair, updated

examples/deepResearch/firecrawlProvider.tsx
```diff
@@ -0,0 +1,46 @@
+import FirecrawlApp, { FirecrawlAppConfig } from "@mendable/firecrawl-js";
+import { gsx } from "gensx";
+
+// Create a context
+export const FirecrawlContext = gsx.createContext<{
+  client?: FirecrawlApp;
+}>({});
+
+// Create the provider
+export const FirecrawlProvider = gsx.Component<FirecrawlAppConfig, never>(
```
EvanBoyle, line 10: We should consider turning this into a package at some point.

EvanBoyle (approved):
Looks good! Should probably use `docs:` instead of `feat:` as features will trigger a new release.

examples/deepResearch/firecrawlProvider.tsx
dereklegenzoff: agreed, i already actually had to duplicate this code from the providers example so even more reason to turn it into a package.

dereklegenzoff:
> Should probably use docs: instead of feat: as features will trigger a new release.

Fair, but now this branch has some updates to `gsx.array` in it, although i could push that to a separate branch if you prefer.

EvanBoyle (approved):
New array impl looks good!

EvanBoyle:
> > Should probably use docs: instead of feat: as features will trigger a new release.
> 
> Fair, but now this branch has some updates to `gsx.array` in it, although i could push that to a separate branch if you prefer.

Makes sense, I left this comment before you made these changes so keeping feat now makes sense.

examples/deepResearch/firecrawlProvider.tsx
jmoseley: We should have a group discussion (that we should have had earlier probably) about the tradeoffs/ergonomics of individual packages (`@gensx/firecrawl`) vs an umbrella package (eg `@gensx/services`). There are maintenance burden tradeoffs I think.

EvanBoyle: I definitely lean towards individual packages even considering the tooling burden, but happy to have that discussion sooner rather than later.

PR #119, opened 2025-01-15
docs: adding the structured outputs doc

dereklegenzoff:
Adds a doc on how to use structured outputs with gensx. shows a basic example as well as a more sophisticated example using zod. 

also renamed the doc to include an "s" at the end to match industry convention of "structured outputs"


docs/src/content/docs/patterns/structured-outputs.mdx
```diff
@@ -0,0 +1,155 @@
...
+        {
+          role: "user",
+          content: prompt,
+        },
+      ]}
+      response_format={zodResponseFormat(ExtractEntitiesSchema, "entities")}
+    >
+      {(response: string) => {
+        try {
+          return ExtractEntitiesSchema.parse(JSON.parse(response));
```
EvanBoyle, line 120: I assume you ran this code and it works, but just want to confirm? 

Would it be worth turning this into a complete example that is checked in that we can link to from this doc?

docs/src/content/docs/patterns/structured-outputs.mdx
```diff
@@ -0,0 +1,155 @@
+---
+title: Structured outputs
+description: Getting structured output from LLMs with GenSX
+sidebar:
+  order: 4
+---
+
+For many workflows, you'll need to get structured output from LLMs. GenSX makes it easy to get structured, type-safe output from LLMs. This guide showcases how to use structured outputs with OpenAI models but you can follow a similar pattern with other LLM providers.
```
EvanBoyle, line 8: Nit: might suggest condensing this. Maybe something like: 

Workflows regularly require getting structured output (JSON) from LLMs. This guide shows how to use structured outputs with OpenAI models but you can follow a similar pattern with other LLM providers.

docs/src/content/docs/patterns/structured-outputs.mdx
```diff
@@ -0,0 +1,155 @@
...
+  places: string[];
+  organizations: string[];
+}
+```
+
+Next, define the component and use the `ChatCompletion` component to get the structured output.
+
+```tsx
+const ExtractEntities = gsx.Component<
+  ExtractEntitiesProps,
```
EvanBoyle, line 34: Nit: the input type isn't actually defined anywhere in the example. Not sure if it needs to be, but it does mean that the example is technically not copy/pastable in current format.

EvanBoyle (approved):
Looks good! Also need an astro.config.js sidebar update to get these changes to show up in the docs nav.

docs/src/content/docs/patterns/structured-outputs.mdx
dereklegenzoff: Yep, i tested it (which was good because it was wrong before i tested it :))

> Would it be worth turning this into a complete example that is checked in that we can link to from this doc?

Good question. A full example with a package.json and everything feels like overkill but maybe we can have somewhere where we have the full code examples from the docs. might be worth chatting about more tomorrow.

docs/src/content/docs/patterns/structured-outputs.mdx
dereklegenzoff: yeah, it's the same input type from the previous example. None of the code blocks are technically runnable in isolation (didn't include the basic imports). 

Riffing on the comment above, maybe the best path is to not worry about code examples being runnable in isolation in the docs but always link out to a fully runnable example.

docs/src/content/docs/patterns/structured-outputs.mdx
dereklegenzoff: fair, pushed an update

docs/src/content/docs/patterns/structured-outputs.mdx
dereklegenzoff: I'm envisioning a `basicExamples` folder within examples where each example from docs is it's own tsx file. then could wire up a pnpm command for each if we wanted. Might mess up the current test setup.

PR #123, opened 2025-01-15
fix: add in missing instruction to examples

dereklegenzoff:
Adding in instructions to set OpenAI API Key in examples


EvanBoyle (approved):

jmoseley (commented):
Missed the reflection example I think.

PR #125, opened 2025-01-15
docs: Update streaming example to use @gensx/openai.

jmoseley:
## Proposed changes

Update streaming example to use @gensx/openai


dereklegenzoff (approved):

examples/streaming/index.tsx
```diff
@@ -12,26 +11,37 @@ async function runStreamingWithChildrenExample() {
...
-        console.log(response);
-      }}
-    </ChatCompletion>,
+    <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
+      <ChatCompletion
+        model="gpt-4o-mini"
+        messages={[{ role: "user", content: prompt }]}
+      >
+        {async (response: string) => {
+          await setTimeout(0);
```
EvanBoyle, line 20: Why did we ever have `await setTimeout(0) in this example? 

Likely a leftover from when this example just returned a string, no? Can we remove it?

EvanBoyle (approved):
Looks good, but we should remove the `setTimeout(0)` assuming they are unnecessary. If they are necessary, we need to have a different conversation lol

examples/streaming/index.tsx
jmoseley: Yeah just left over from needing proof that async worked.

PR #128, opened 2025-01-16
docs: adding basic concepts doc and refactoring quickstart

dereklegenzoff:
- Updates the quickstart
- adds a basic concepts doc that explains components, composition, execution, providers, etc

### DIFF
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -110,13 +110,59 @@ Components are executed using `gsx.execute()`, which processes the JSX tree from
 const result = await gsx.execute(<MyComponent input="some data" />);
 ```
 
+You'll often use `gsx.execute()` to run a top level workflow as shown above, but you can also use it to run a sub-workflow within a component. This allows you to work with a component in TypeScript without having to drop into the JSX layer. This is especially useful for more complex operations like map/reduce:
+
+```tsx
+const AnalyzeReviews = gsx.Component<AnalyzeReviewsProps, AnalyzeReviewsOutput>(
+  async ({ reviews }) => {
+    // Map: Extract topics from each review in parallel
+    const topics = await Promise.all(
+      reviews.map((review) => gsx.execute(<GetTopic review={review} />)),
+    );
+    // Reduce: Count frequency of each topic
+    return topics.reduce(
+      (counts, topic) => ({ ...counts, [topic]: (counts[topic] || 0) + 1 }),
+      {},
+    );
+  },
+);
+```
+
 Components can be tested and evaluated in isolation, making it easy to debug and verify individual steps of your workflow. This is particularly valuable when building complex LLM applications that need robust evaluation.
 
 Rather than having to run an entire workflow to test a change to a single component, you can test just that component, dramatically speeding up your dev loop. This isolation also makes unit testing more manageable, as you can create specific test cases without having to worry about the rest of the workflow.
 
+## Contexts
+
+Contexts provide a way to share values across components without explicitly passing them through props. They're similar to React's Context system but adapted for GenSX workflows. Contexts are particularly useful for:
+
+- Sharing configuration across multiple components
+- Providing dependencies to deeply nested components
+- Managing state that needs to be accessed by multiple components
+
+Here's how to create and use a context:
+
+```tsx
+// Create a context with a default value
+const UserContext = gsx.createContext({ name: "default", age: 0 });
+
+// Use the context in a component
+const UserGreeting = gsx.Component(() => {
+  const user = gsx.useContext(UserContext);
+  return `Hello, ${user.name}!`;
+});
+
+// Provide a value to the context
+const result = await gsx.execute(
+  <UserContext.Provider value={{ name: "John", age: 30 }}>
+    <UserGreeting />
+  </UserContext.Provider>,
+);
+```
+
 ## Providers
 
-Providers are special components that manage configuration and dependencies for your workflow. They're similar to React's Context providers - they just pass configuration down to child components.
+Providers are a specialized use of contexts that focus on managing configuration and dependencies for your workflow. They're built using the context system described above, but provide a more focused API for configuration management.
 
 The main provider available today is the `OpenAIProvider`, which manages your OpenAI API key:
 
@@ -131,38 +177,9 @@ const result = await gsx.execute(
 );
 ```
 
-Providers allow you to:
+Providers offer several benefits:
 
-- Configure dependencies like API keys and other settings
-- Avoid passing configuration through multiple levels of components
-- Keep components pure and reusable by separating configuration from logic
-
-## Contexts
-
-Contexts provide a way to share values across components without explicitly passing them through props. They're similar to React's Context system but adapted for GenSX workflows. Contexts are particularly useful for:
-
-- Sharing configuration across multiple components
-- Providing dependencies to deeply nested components
-- Managing state that needs to be accessed by multiple components
-
-Here's how to create and use a context:
-
-```tsx
-import { createContext, useContext } from "gensx";
-
-// Create a context with a default value
-const UserContext = createContext({ name: "default", age: 0 });
-
-// Use the context in a component
-const UserGreeting = gsx.Component(() => {
-  const user = useContext(UserContext);
-  return `Hello, ${user.name}!`;
-});
-
-// Provide a value to the context
-const result = await gsx.execute(
-  <UserContext.Provider value={{ name: "John", age: 30 }}>
-    <UserGreeting />
-  </UserContext.Provider>,
-);
-```
+- Manage complex configuration like API clients and settings
+- Provide a simpler, more focused API than raw contexts
+- Keep configuration separate from business logic
+- Enable dependency injection for testing and development
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -1,0 +1,116 @@
+---
+title: Basic Concepts
+description: Understanding the core concepts of GenSX
+---
+
+GenSX is a workflow engine designed for building complex LLM applications. It's built around functional, reusable components that are composed using JSX to create and orchestrate complicated workflows.
+
+## Components
+
+Components are the building blocks of GenSX applications; they're pure TypeScript functions that:
+
+- Accept props as input
+- Return a single output or another component
+- Have zero side effects
+- Don't depend on external state
+- Are fully typed using TypeScript
+
+Here's an example of a simple component:
+
+```tsx
+interface GreetingProps {
+  name: string;
+}
+type GreetingOutput = string;
+
+const Greeting = gsx.Component<GreetingProps, GreetingOutput>(
+  async ({ name }) => {
+    return `Hello, ${name}!`;
+  },
+);
+```
+
+Components can also return other components using JSX:
+
+```tsx
+const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
+  return (
+    <ChatCompletion
+      model="gpt-4o-mini"
+      messages={[
+        {
+          role: "system",
+          content: "You are a friendly assistant that greets people warmly.",
+        },
+        { role: "user", content: `Write a greeting for ${name}.` },
+      ]}
+    />
+  );
+});
+```
+
+## JSX and Data Flow
+
+GenSX uses JSX to compose components into workflows. Data flows from parent to child through props allowing you to nest components and pass data between them. When components are combined, they will return the value of the innermost component.
+
+```tsx
+// Parent receives output from ChildComponent through the child function
+<ParentComponent input="some data">
+  {(parentResult) => <ChildComponent data={parentResult} />}
+</ParentComponent>
+```
+
+Unlike React's concurrent rendering model, GenSX components execute sequentially from top to bottom. Each component completes before its children begin.
+
+## Component Types
+
+GenSX provides two main types of components:
+
+1. **Regular Components** (`gsx.Component`) - For standard synchronous or asynchronous operations
+2. **Streaming Components** (`gsx.StreamingComponent`) - For handling streaming responses from LLMs
+
+For streaming components, you can enable streaming by passing the `stream` prop and consuming the result as an async iterator:
+
+```tsx
+const stream = await gsx.execute(
+  <MyStreamingComponent input="some data" stream={true} />,
+);
+
+// Process the streaming response
+for await (const chunk of stream) {
+  process.stdout.write(chunk);
+}
+```
+
+## Executing Components
+
+Components are executed using `gsx.execute()`, which processes the JSX tree from top to bottom:
+
+```tsx
+const result = await gsx.execute(<MyComponent input="some data" />);
+```
+
+Components can be tested and evaluated in isolation, making it easy to debug and verify individual steps of your workflow. This is particularly valuable when building complex LLM applications that need robust evaluation.
+
+## Providers
+
+Providers are special components that manage configuration and dependencies for your workflow. They're similar to React's Context providers but simpler - they just pass configuration down to child components.
+
+The main provider available today is the `OpenAIProvider`, which manages your OpenAI API key:
+
+```tsx
+const result = await gsx.execute(
+  <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
+    <ChatCompletion
+      model="gpt-4o-mini"
+      messages={[{ role: "user", content: "Hello!" }]}
+    />
+  </OpenAIProvider>,
+);
+```
+
+Providers allow you to:
+
+- Configure dependencies like API keys and other settings
+- Avoid passing configuration through multiple levels of components
+- Keep components pure and reusable by separating configuration from logic
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -50,7 +50,7 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
 });
 ```
 
-Components can return structured objects containing other components. The pattern also works with array operations like `map` to process items in parallel.
+Components can also return structured objects containing other components. The pattern works with array operations like `map` for processing items in parallel.
 
 ```tsx
 const AnalyzePosts = gsx.Component<AnalyzePostsProps, AnalyzePostsOutput>(
@@ -86,10 +86,10 @@ Unlike React's concurrent rendering model, GenSX evaluates your workflow as a de
 
 GenSX provides two main types of components:
 
-1. **Components** (`gsx.Component`) - For standard synchronous or asynchronous operations
-2. **Streaming Components** (`gsx.StreamingComponent`) - For handling streaming responses from LLMs
+1. **Components** (`gsx.Component`) - Components that produce values and can handle both synchronous and asynchronous operations
+2. **Streaming Components** (`gsx.StreamingComponent`) - Components designed to handle responses from LLMs, working in both streaming and non-streaming contexts
 
-For streaming components, you can enable streaming by passing the `stream` prop and consuming the result as an async iterator:
+The power of a `StreamingComponent` is that a single implementation can be used interchangeably in both streaming and non-streaming contexts without any code changes - you just toggle the `stream` prop:
 
 ```tsx
 const stream = await gsx.execute<Streamable>(
@@ -112,9 +112,7 @@ const result = await gsx.execute(<MyComponent input="some data" />);
 
 Components can be tested and evaluated in isolation, making it easy to debug and verify individual steps of your workflow. This is particularly valuable when building complex LLM applications that need robust evaluation.
 
-## Contexts
-
-TODO - explain contexts
+Rather than having to run an entire workflow to test a change to a single component, you can test just that component, dramatically speeding up your dev loop. This isolation also makes unit testing more manageable, as you can create specific test cases without having to worry about the rest of the workflow.
 
 ## Providers
 
@@ -138,3 +136,33 @@ Providers allow you to:
 - Configure dependencies like API keys and other settings
 - Avoid passing configuration through multiple levels of components
 - Keep components pure and reusable by separating configuration from logic
+
+## Contexts
+
+Contexts provide a way to share values across components without explicitly passing them through props. They're similar to React's Context system but adapted for GenSX workflows. Contexts are particularly useful for:
+
+- Sharing configuration across multiple components
+- Providing dependencies to deeply nested components
+- Managing state that needs to be accessed by multiple components
+
+Here's how to create and use a context:
+
+```tsx
+import { createContext, useContext } from "gensx";
+
+// Create a context with a default value
+const UserContext = createContext({ name: "default", age: 0 });
+
+// Use the context in a component
+const UserGreeting = gsx.Component(() => {
+  const user = useContext(UserContext);
+  return `Hello, ${user.name}!`;
+});
+
+// Provide a value to the context
+const result = await gsx.execute(
+  <UserContext.Provider value={{ name: "John", age: 30 }}>
+    <UserGreeting />
+  </UserContext.Provider>,
+);
+```
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -116,8 +116,8 @@ You'll often use `gsx.execute()` to run a top level workflow as shown above, but
 const AnalyzeReviews = gsx.Component<AnalyzeReviewsProps, AnalyzeReviewsOutput>(
   async ({ reviews }) => {
     // Map: Extract topics from each review in parallel
-    const topics = await Promise.all(
-      reviews.map((review) => gsx.execute(<GetTopic review={review} />)),
+    const topics = await gsx.execute<GetTopicOutput[]>(
+      reviews.map((r) => <GetTopic review={r} />),
     );
     // Reduce: Count frequency of each topic
     return topics.reduce(
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -50,7 +50,7 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
 });
 ```
 
-Components can also return arrays and structured objects containing other components. At runtime the framework takes care of resolving all components into a singular output. This pattern works with array operations like `map` for processing items in parallel.
+Components can also return arrays and structured objects containing other components. The pattern also works with array operations like `map` for processing items in parallel.
 
 ```tsx
 const AnalyzePosts = gsx.Component<AnalyzePostsProps, AnalyzePostsOutput>(
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -14,7 +14,7 @@ Components are the building blocks of GenSX applications; they're pure TypeScrip
 - Don't depend on global state
 - Are strongly typed using TypeScript
 
-Components can return a primitive, an object, or other components. These outputs can then be used by other components or returned as the result of the workflow.
+Components can return a primitive, an object, or other components. Their outputs can be used by other components or returned as the result of the workflow.
 
 Here's an example of a simple component:
 
@@ -31,7 +31,7 @@ const Greeting = gsx.Component<GreetingProps, GreetingOutput>(
 );
 ```
 
-Component outputs can be composed of other components using JSX:
+Here's an example of a component that returns another component:
 
 ```tsx
 const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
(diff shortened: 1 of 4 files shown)


docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
+---
+title: Basic Concepts
+description: Understanding the core concepts of GenSX
+---
+
+GenSX is a workflow engine designed for building complex LLM applications. It's built around functional, reusable components that are composed using JSX to create and orchestrate complicated workflows.
```
EvanBoyle, line 6: I kinda like the simple/complex dichotomy used in the overview page: 

> GenSX is a simple typescript framework for building complex LLM applications.

We use complicated and complex in this intro. Overemphasizing complexity sends the wrong message. Key points we want to drill home throughout our docs (not necessarily all at once in this intro): 

- simple framework
- make the complex things easy
- composable
- functional and reusable
-

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+## Components
+
+Components are the building blocks of GenSX applications; they're pure TypeScript functions that:
```
EvanBoyle, line 10: The number one most important thing to drill home is that a component is a function that produces a value to be used by other components and parts of the workflow. That value could be LLM output in the form of a string, structured output, a stream, or results from an API or external datastore like a vector database query.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+- Accept props as input
+- Return a single output or another component
```
EvanBoyle, line 13: I might say: 

"Produces an output that can be used by another component"
Perhaps separately or in parenthesis add "(and outputs can be composed of other components!)"

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+- Have zero side effects
```
EvanBoyle, line 14: Not true, you can do whatever you want in a component, write to S3, update a db row, etc.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+- Don't depend on external state
+- Are fully typed using TypeScript
```
EvanBoyle, line 16: maybe replace `fully` with `strongly`

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+GenSX is a workflow engine designed for building complex LLM applications. It's built around functional, reusable components that are composed using JSX to create and orchestrate complicated workflows.
+
+## Components
+
+Components are the building blocks of GenSX applications; they're pure TypeScript functions that:
+
+- Accept props as input
+- Return a single output or another component
+- Have zero side effects
+- Don't depend on external state
```
EvanBoyle, line 15: Many components will depend on external state but not depending on global state is an explicit design goal.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+type GreetingOutput = string;
+
+const Greeting = gsx.Component<GreetingProps, GreetingOutput>(
+  async ({ name }) => {
+    return `Hello, ${name}!`;
+  },
+);
+```
+
+Components can also return other components using JSX:
```
EvanBoyle, line 33: I would change this to "Component outputs can be composed from other components using JSX".

Also, do you think it is worth talking about arrays and objects? Maybe an extra example right after this one: 

```jsx
const AnalyzeHNPosts = gsx.Component<AnalyzeHNPostsProps, AnalyzeHNPostsOutput>(
  ({ stories }) => ({
    analyses: stories.map((story) => ({
      summary: <PostSummarizer story={story} />,
      commentAnalysis: (
        <CommentsAnalyzer postId={story.id} comments={story.comments} />
      ),
    })),
  }),
);
```

I particularly like this example because it shows composition with both array.map and object + jsx syntax. quite powerful and meaty in just a few lines of code. This is not standard JSX, but something we've added on. So users won't know it exists if we don't talk about it.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+        { role: "user", content: `Write a greeting for ${name}.` },
+      ]}
+    />
+  );
+});
+```
+
+## JSX and Data Flow
+
+GenSX uses JSX to compose components into workflows. Data flows from parent to child through props allowing you to nest components and pass data between them. When components are combined, they will return the value of the innermost component.
```
EvanBoyle, line 54: > Data flows from parent to child through props

I feel like this is not quite a precise as it needs to be. 

Yes props is the mechanism for receiving dependencies, but outputs and child function nesting is the mechanism for passing them. Don't want to overcomplicate here, but maybe we can make a small adjustment.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+</ParentComponent>
+```
+
+Unlike React's concurrent rendering model, GenSX components execute sequentially from top to bottom. Each component completes before its children begin.
+
+## Component Types
+
+GenSX provides two main types of components:
+
+1. **Regular Components** (`gsx.Component`) - For standard synchronous or asynchronous operations
```
EvanBoyle, line 69: I'd drop the Regular and just refer to them as `Components`

EvanBoyle: Also they always produce prompt values. They can return promises and the framework auto-resolves them when they are accessed via nesting. Don't necessarily need to mention all of that here but just expanding on my mental model in case it is helpful.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+2. **Streaming Components** (`gsx.StreamingComponent`) - For handling streaming responses from LLMs
```
EvanBoyle, line 70: The interesting thing here is not being able to handle streams, but being able to write a single component that can be used interchangeably for streaming and prompt output without any major code changes.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+GenSX uses JSX to compose components into workflows. Data flows from parent to child through props allowing you to nest components and pass data between them. When components are combined, they will return the value of the innermost component.
+
+```tsx
+// Parent receives output from ChildComponent through the child function
+<ParentComponent input="some data">
+  {(parentResult) => <ChildComponent data={parentResult} />}
+</ParentComponent>
+```
+
+Unlike React's concurrent rendering model, GenSX components execute sequentially from top to bottom. Each component completes before its children begin.
```
EvanBoyle, line 63: Another tidbit that might be worth mentioning somewhere is that the execution is parallel by default. We build a dependency graph from your workflow definition, and execute it in parallel without you having to worry about any of it.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+// Process the streaming response
+for await (const chunk of stream) {
+  process.stdout.write(chunk);
+}
+```
+
+## Executing Components
+
+Components are executed using `gsx.execute()`, which processes the JSX tree from top to bottom:
```
EvanBoyle, line 87: `gsx.execute` has two use cases: 

1. Kick off a top level workflow (ie a component) and await the result/output
2. kick off a sub-workflow within a component in typescript vs dropping into the JSX layer. Useful for complex map/reduce/filter operations.

EvanBoyle: Not sure if that is necessary to mention here, but that is my mental model and want to make sure that at least we don't contradict that. And we need to talk about this pattern somewhere in our docs eventually.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+```tsx
+const result = await gsx.execute(<MyComponent input="some data" />);
+```
+
+Components can be tested and evaluated in isolation, making it easy to debug and verify individual steps of your workflow. This is particularly valuable when building complex LLM applications that need robust evaluation.
```
EvanBoyle, line 93: Maybe we can make this a little bit more relatable. Being able to mock up inputs for a single component and test it in isolation is a wayyyyy faster dev loop that running an entire 20 step workflow over and over again that takes two minutes, not to mention cheaper as well!

This section is straightforward and compelling, but I think it undersells the impact a little bit.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+## Providers
+
+Providers are special components that manage configuration and dependencies for your workflow. They're similar to React's Context providers but simpler - they just pass configuration down to child components.
```
EvanBoyle, line 97: > but simpler

The APIs are pretty close. Are ours simpler? I think they let you express more or less the same things.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+    />
+  </OpenAIProvider>,
+);
+```
+
+Providers allow you to:
+
+- Configure dependencies like API keys and other settings
+- Avoid passing configuration through multiple levels of components
+- Keep components pure and reusable by separating configuration from logic
```
EvanBoyle, line 116: I like this section. We also have the full blown context API. This section doesn't show how to use it to pass around configuration. Should we have an example here?

EvanBoyle (commented):
This is looking really good! The docs are both information packed, but extremely easy reading which is a really good thing. 

Most of the comments I left are focused on the precision of the language we use. I want to be precise about these concepts, but keep the straightforward simple and easy to grok explanations. might take a rev or two to strike a happy medium.

docs/src/content/docs/quickstart.mdx
```diff
@@ -54,22 +60,156 @@ const Respond = gsx.Component<RespondProps, RespondOutput>(
...
 );
 
-const App = gsx.Component<Record<string, never>, string>(async () => {
-  return await gsx.execute(
-    <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
-      <Respond userInput="Hi there! Say 'Hello, World!' and nothing else." />
-    </OpenAIProvider>,
+const result = await gsx.execute(
+  <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
+    <Respond userInput="Hi there! Say 'Hello, World!' and nothing else." />
```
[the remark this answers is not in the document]
dereklegenzoff: interesting point. I actually like having the wrapper because I think it illustrates the most common pattern people will use for calling LLMs (wrapping the `<ChatCompletion>` in a component to encapsulate some business logic). I'd prefer to show users the most common pattern rather than the simplest pattern first. Plus, in terms of the starter template, this is a better jumping off point. 

I do agree that in it's current form, the `Respond` doesn't add any value. Maybe we should make some simple update like having the component fetch/format the current date and add that to the system message and maybe renaming it.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good point

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good call

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good catch

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: oh yeah i really like both the proposed text and the example. will make the updates.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good point, i'll make sure to call that out

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: well said, i'll add that in. I had been thinking this sentence might make someone think that we wait for the entire prior step to finish

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good point, this crossed my mind. will just want to make sure that the pattern in the example is actually something we'd want someone to use

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: fair, i'll remove that.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good point, i'll add a section now that we merged that PR.

dereklegenzoff:
> This is looking really good! The docs are both information packed, but extremely easy reading which is a really good thing.

Thanks! Glad to hear the articles scored well in the human evals for depth and readability 😉 

> Most of the comments I left are focused on the precision of the language we use. I want to be precise about these concepts, but keep the straightforward simple and easy to grok explanations. might take a rev or two to strike a happy medium.

Thanks for all the feedback! Appreciate the help making things precise. Once we nail the verbiage a few times, it gets much easier the following times so definitely worth the effort in getting it right.

docs/src/content/docs/quickstart.mdx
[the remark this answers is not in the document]
EvanBoyle: I wouldn't make the example any more complicated. Starting with the simplest hello world, just a few lines to call an LLM is a good thing! That is what people will expect. Don't want to overload them with complexity on the first screen.

docs/src/content/docs/basic-concepts.mdx
jmoseley: I agree, I think we should be clear about the necessity of the child function, and how that works.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+GenSX provides two main types of components:
+
+1. **Regular Components** (`gsx.Component`) - For standard synchronous or asynchronous operations
+2. **Streaming Components** (`gsx.StreamingComponent`) - For handling streaming responses from LLMs
+
+For streaming components, you can enable streaming by passing the `stream` prop and consuming the result as an async iterator:
+
+```tsx
+const stream = await gsx.execute(
```
jmoseley, line 75: We should be careful to always include the type params.

```suggestion
const stream = await gsx.execute<Streamable>(
```

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: Sound feedback. Updated to this

```
GenSX is a simple typescript framework for building complex LLM applications. It's built around functional, reusable components that are composed using JSX to create and orchestrate workflows.
```

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: removed this bullet

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: updated "external" to "global"

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: Added this - feels like i could do a better job explaining it but the words are escaping me

---

Components can also produce objects that contain other components:

```tsx
const AnalyzePosts = gsx.Component<AnalyzePostsProps, AnalyzePostsOutput>(
  ({ posts }) => ({
    analyses: posts.map((post) => ({
      summary: <PostSummarizer post={post} />,
      commentAnalysis: (
        <CommentsAnalyzer postId={post.id} comments={post.comments} />
      ),
    })),
  }),
);
```

In this example, `AnalyzePosts` will produce an object that contains the `summaries` and `commentAnalysis` for each post.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good catch

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: Improved it to this

---

Components can return structured objects containing other components. The pattern also works with array operations like `map` to process items in parallel.

```tsx
const AnalyzePosts = gsx.Component<AnalyzePostsProps, AnalyzePostsOutput>(
  ({ posts }) => ({
    analyses: posts.map((post) => ({
      summary: <PostSummarizer post={post} />,
      commentAnalysis: (
        <CommentsAnalyzer postId={post.id} comments={post.comments} />
      ),
    })),
  }),
);
```

In this example, `AnalyzePosts` will produce an array of objects that contains the `summaries` and `commentAnalysis` for each post.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: Good feedback. Here's my latest draft of this: 
---
GenSX uses JSX to compose components into workflows. The output of a parent component can be passed to a child component through a child function. This pattern enables you to create chains of components where each step's output feeds into the next component's input.

When components are combined, they will return the value of the innermost component.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: planning to update to this

---

Unlike React's concurrent rendering model, GenSX evaluates your workflow as a dependency graph. Components execute in parallel whenever possible, while automatically ensuring that all required inputs are available before a component starts executing.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: good feedback. revved on the intro to get it to this

---

Components are the building blocks of GenSX applications; they're pure TypeScript functions that:

- Accept props as input
- Produce an output
- Don't depend on global state
- Are strongly typed using TypeScript

Components can return a primitive, an object, or other components. These outputs can then be used by other components or returned as the result of the workflow.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: landed on this

---

GenSX provides two main types of components:

1. **Components** (`gsx.Component`) - Standard components for synchronous and asynchronous operations
2. **Streaming Components** (`gsx.StreamingComponent`) - Components designed to handle responses from LLMs, working in both streaming and non-streaming contexts

The power of a `StreamingComponent` is that a single implementation can be used interchangeably in both streaming and non-streaming contexts without any code changes - you just toggle the `stream` prop:

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: fair. rather than changing what's there, i'm thinking i'll add on a couple of sentences

--- 

Rather than having to run an entire workflow to test a change to a single component, you can test just that component, dramatically speeding up your dev loop. This isolation also makes unit testing more manageable, as you can create specific test cases without having to worry about the rest of the workflow.

docs/src/content/docs/basic-concepts.mdx
EvanBoyle: For this: 

> Components can return a primitive, an object, or other components. These outputs can then be used by other components or returned as the result of the workflow.

What would you think about: 

```
Components can not only return data, but also other components. This includes arrays of components, and even objects where arbitrary keys are components as well. At runtime the framework takes care of resolving all components into a singular output: 

... example showing object/array, etc
```

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -49,9 +50,28 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
 });
 ```
 
+Components can return structured objects containing other components. The pattern also works with array operations like `map` to process items in parallel.
```
EvanBoyle: "Components can return arrays and structured objects containing other ocmponents"

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -49,9 +50,28 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
...
+      summary: <PostSummarizer post={post} />,
+      commentAnalysis: (
+        <CommentsAnalyzer postId={post.id} comments={post.comments} />
+      ),
+    })),
+  }),
+);
+```
+
+In this example, `AnalyzePosts` will produce an array of objects that contains the `summaries` and `commentAnalysis` for each post.
```
EvanBoyle, line 68: Mention that the framework handles resolving all of the nested components and collecting them into a final value. No extra work required.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -49,9 +50,28 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
...
+
 ## JSX and Data Flow
 
-GenSX uses JSX to compose components into workflows. Data flows from parent to child through props allowing you to nest components and pass data between them. When components are combined, they will return the value of the innermost component.
+GenSX uses JSX to compose components into workflows. The output of a parent component can be passed to a child component through a child function. This pattern enables you to create chains of components where each step's output feeds into the next component's input.
+
+When components are combined, they will return the value of the innermost component.
```
EvanBoyle, line 74: Reading this again, this is mostly true. But the precise version is that a tree of components composed via child nesting returns either (1) the deepest nested component: 

```tsx
<A><B><C/></B></> // returns C
```

or it returns the final return value: 

```tsx
<A>
{a => 
  <B>
  { b => 
    return {a, b, c: <C/>}
  {
  </B>
 }
<A/> // returns C
```

In the second example, C is technically the most deeply nested component but that example returns `{a, b, c}`. Not sure if we need to go into that much detail, but that is the precise definition and you can do with it what you please

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -60,19 +80,19 @@ GenSX uses JSX to compose components into workflows. Data flows from parent to c
 </ParentComponent>
 ```
 
-Unlike React's concurrent rendering model, GenSX components execute sequentially from top to bottom. Each component completes before its children begin.
+Unlike React's concurrent rendering model, GenSX evaluates your workflow as a dependency graph. Components execute in parallel whenever possible, while automatically ensuring that all required inputs are available before a component starts executing.
```
EvanBoyle, line 83: Do you think we need to talk about trees vs graphs at all here? https://gensx.dev/why-jsx/#graphs-dags-and-trees

Maybe something like: 

Just because GenSX uses trees doesn't mean it can't represent cycles and everything else you would expect from graphs and agents. To learn more see (https://gensx.dev/why-jsx/#graphs-dags-and-trees)...

EvanBoyle (commented):
few more comments

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -86,10 +86,10 @@ Unlike React's concurrent rendering model, GenSX evaluates your workflow as a de
 
 GenSX provides two main types of components:
 
-1. **Components** (`gsx.Component`) - For standard synchronous or asynchronous operations
-2. **Streaming Components** (`gsx.StreamingComponent`) - For handling streaming responses from LLMs
+1. **Components** (`gsx.Component`) - Components that produce values and can handle both synchronous and asynchronous operations
```
EvanBoyle, line 89: nit: `that produce an output value` -- still a singular output whether it be an array, object, string, number, etc.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -86,10 +86,10 @@ Unlike React's concurrent rendering model, GenSX evaluates your workflow as a de
...
+2. **Streaming Components** (`gsx.StreamingComponent`) - Components designed to handle responses from LLMs, working in both streaming and non-streaming contexts
 
-For streaming components, you can enable streaming by passing the `stream` prop and consuming the result as an async iterator:
+The power of a `StreamingComponent` is that a single implementation can be used interchangeably in both streaming and non-streaming contexts without any code changes - you just toggle the `stream` prop:
```
EvanBoyle, line 92: 👨‍🍳

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -138,3 +136,33 @@ Providers allow you to:
...
+  return `Hello, ${user.name}!`;
+});
+
+// Provide a value to the context
+const result = await gsx.execute(
+  <UserContext.Provider value={{ name: "John", age: 30 }}>
+    <UserGreeting />
+  </UserContext.Provider>,
+);
+```
```
EvanBoyle, line 161: Might be worth mentioning something like: 

```
Contexts are a useful way to pass configuration without prop drilling. However, they do make your components less reusable so it is recommended that you use them sparingly. 
```

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -110,32 +110,27 @@ Components are executed using `gsx.execute()`, which processes the JSX tree from
...
-Components can be tested and evaluated in isolation, making it easy to debug and verify individual steps of your workflow. This is particularly valuable when building complex LLM applications that need robust evaluation.
-
-Rather than having to run an entire workflow to test a change to a single component, you can test just that component, dramatically speeding up your dev loop. This isolation also makes unit testing more manageable, as you can create specific test cases without having to worry about the rest of the workflow.
-
-## Providers
-
-Providers are special components that manage configuration and dependencies for your workflow. They're similar to React's Context providers - they just pass configuration down to child components.
-
-The main provider available today is the `OpenAIProvider`, which manages your OpenAI API key:
+You'll often use `gsx.execute()` to run a top level workflow as shown above, but you can also use it to run a sub-workflow within a component. This allows you to work with a component in TypeScript without having to drop into the JSX layer. This is especially useful for more complex operations like map/reduce:
```
EvanBoyle, line 113: should also mention here that `gsx.execute` within a component maintains all current context and parent/child relationships with the current component and the rest of the workflow.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -50,7 +50,7 @@ const AIGreeting = gsx.Component<GreetingProps, string>(async ({ name }) => {
 });
 ```
 
-Components can also return structured objects containing other components. The pattern works with array operations like `map` for processing items in parallel.
+Components can also return structured objects containing. The pattern works with array operations like `map` for processing items in parallel.
```
EvanBoyle, line 53: Dropped a word at the end of sentence one.

EvanBoyle (approved):
🚢 looks great overall! I left a few comments for your consideration, but I think were in a pretty good place over all.

docs/src/content/docs/basic-concepts.mdx
```diff
@@ -104,7 +104,7 @@ for await (const chunk of stream) {
 
 ## Executing Components
 
-Components are executed using `gsx.execute()`, which processes the JSX tree from top to bottom:
+Components are executed using `gsx.execute()`, which processes the JSX tree from top to bottom while executing components in parallel as much as possible.
```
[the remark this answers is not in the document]
dereklegenzoff: i like it; updated

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: fair, updated

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: ooh yeah i like that; will add something in

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: i like it. Adding it in.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: added

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: updated

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: added

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: Here's what i landed on. Not perfect but workable.

---

While GenSX uses a [tree-based structure with JSX](https://gensx.dev/why-jsx/#graphs-dags-and-trees), it can represent everything you'd expect from graph-based workflows including cycles and agentic patterns

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: interesting. very well explained. I think i'll leave the current phrasing because it will give folks an accurate mental model and i'm hesitant to add more complexity.

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: made some updates to incorporate a lot of this verbiage

PR #130, opened 2025-01-16
feat: Support multiple OpenAI Client contexts simultaneously

jmoseley:
## Proposed changes

Improve the pattern for creating OpenAI client contexts, to better support a workflow that uses multiple OpenAI compatible APIs, while supporting the sharing model. See the blog writer example for how this might work in the wild.

For a shared component, there is a `ChatCompletion` component that has a `context` prop, to make it configurable.

### DIFF
--- a/examples/blogWriter/perplexity.tsx
+++ b/examples/blogWriter/perplexity.tsx
@@ -1,0 +1,15 @@
+import { createOpenAIClientContext } from "@gensx/openai";
+
+const { Provider, ChatCompletion } = createOpenAIClientContext<
+  | "llama-3.1-sonar-small-128k-online"
+  | "llama-3.1-sonar-large-128k-online"
+  | "llama-3.1-sonar-huge-128k-online"
+>(
+  {
+    apiKey: process.env.PERPLEXITY_API_KEY,
+    baseURL: "https://api.perplexity.ai/chat/completions",
+  },
+  "Perplexity",
+);
+
+export { Provider, ChatCompletion };
--- a/packages/gensx-openai/src/index.tsx
+++ b/packages/gensx-openai/src/index.tsx
@@ -8,58 +8,77 @@ import {
 } from "openai/resources/index.mjs";
 import { Stream } from "openai/streaming";
 
-// Create a context for OpenAI
-export const OpenAIContext = gsx.createContext<{
-  client?: OpenAI;
-}>({});
+/**
+ * Create a context for an OpenAI client and bind it to a ChatCompletion component. This allows you to have
+ * multiple OpenAI clients that target different OpenAI-compatible API endpoints.
+ *
+ * @param args - The client options
+ * @returns The OpenAI context and the ChatCompletion component
+ */
+export const createOpenAIClientContext = <
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-parameters
+  ModelName extends string = ChatCompletionCreateParams["model"],
+>(
+  args: ClientOptions,
+  namePrefix: string,
+) => {
+  const client = new OpenAI(args);
 
-export const OpenAIProvider = gsx.Component<ClientOptions, never>(
-  "OpenAIProvider",
-  (args) => {
-    const client = new OpenAI(args);
-    return <OpenAIContext.Provider value={{ client }} />;
-  },
-);
+  const OpenAIClientContext = gsx.createContext(client);
 
-// Create a component for chat completions
-export const ChatCompletion = gsx.StreamComponent<ChatCompletionCreateParams>(
-  "ChatCompletion",
-  async (props) => {
-    const context = gsx.useContext(OpenAIContext);
-
-    if (!context.client) {
-      throw new Error(
-        "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
-      );
+  const BoundChatCompletion = gsx.StreamComponent<
+    ChatCompletionCreateParams & {
+      model: ModelName;
     }
+  >(`${namePrefix}ChatCompletion`, (props) => {
+    const { stream, ...rest } = props;
+    return (
+      <ChatCompletion
+        context={OpenAIClientContext}
+        stream={stream ?? false}
+        {...rest}
+      />
+    );
+  });
 
-    if (props.stream) {
-      const stream = await context.client.chat.completions.create(props);
+  function OpenAIClientProvider(_: gsx.Args<{}, never>) {
+    return <OpenAIClientContext.Provider value={client} />;
+  }
 
-      async function* generateTokens(): AsyncGenerator<
-        string,
-        void,
-        undefined
-      > {
-        for await (const chunk of stream as Stream<ChatCompletionChunk>) {
-          const content = chunk.choices[0]?.delta?.content;
-          if (content) {
-            yield content;
-          }
+  return {
+    Provider: OpenAIClientProvider,
+    ChatCompletion: BoundChatCompletion,
+  };
+};
+
+export const ChatCompletion = gsx.StreamComponent<
+  ChatCompletionCreateParams & { context: gsx.Context<OpenAI> }
+>("ChatCompletion", async (props) => {
+  const { context, ...createParams } = props;
+  const client = gsx.useContext(context);
+
+  if (props.stream) {
+    const stream = await client.chat.completions.create(createParams);
+
+    async function* generateTokens(): AsyncGenerator<string, void, undefined> {
+      for await (const chunk of stream as Stream<ChatCompletionChunk>) {
+        const content = chunk.choices[0]?.delta?.content;
+        if (content) {
+          yield content;
         }
       }
-
-      const streamable: Streamable = generateTokens();
-      return streamable;
-    } else {
-      const response = await context.client.chat.completions.create(props);
-      const content = response.choices[0]?.message?.content ?? "";
-
-      function* generateTokens() {
-        yield content;
-      }
-
-      return generateTokens();
     }
-  },
-);
+
+    const streamable: Streamable = generateTokens();
+    return streamable;
+  } else {
+    const response = await client.chat.completions.create(props);
+    const content = response.choices[0]?.message?.content ?? "";
+
+    function* generateTokens() {
+      yield content;
+    }
+
+    return generateTokens();
+  }
+});
--- a/examples/blogWriter/blogWriter.tsx
+++ b/examples/blogWriter/blogWriter.tsx
@@ -1,5 +1,13 @@
-import { ChatCompletion, OpenAIProvider } from "@gensx/openai";
 import { gsx } from "gensx";
+
+import {
+  ChatCompletion as OpenAIChatCompletion,
+  Provider as OpenAIProvider,
+} from "./openai.js";
+import {
+  ChatCompletion as PplxChatCompletion,
+  Provider as PplxProvider,
+} from "./perplexity.js";
 
 interface LLMResearchBrainstormProps {
   prompt: string;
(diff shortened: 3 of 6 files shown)


examples/blogWriter/perplexity.tsx
```diff
@@ -0,0 +1,15 @@
...
+const { Provider, ChatCompletion } = createOpenAIClientContext<
+  | "llama-3.1-sonar-small-128k-online"
+  | "llama-3.1-sonar-large-128k-online"
+  | "llama-3.1-sonar-huge-128k-online"
+>(
+  {
+    apiKey: process.env.PERPLEXITY_API_KEY,
+    baseURL: "https://api.perplexity.ai/chat/completions",
+  },
+  "Perplexity",
```
dereklegenzoff, line 12: is it weird that in gsx.Component the name goes first but it's second here?

packages/gensx-openai/src/index.tsx
```diff
@@ -8,58 +8,77 @@ import {
...
-        for await (const chunk of stream as Stream<ChatCompletionChunk>) {
-          const content = chunk.choices[0]?.delta?.content;
-          if (content) {
-            yield content;
-          }
-        }
-      }
+  return {
+    Provider: OpenAIClientProvider,
+    ChatCompletion: BoundChatCompletion,
```
EvanBoyle, line 50: Feels like we should give an output type to this thing, something like `OpenAICompatibleContext`. Then component authors can write reusable components that accept a context: 

```tsx
interface WebResearchProps {
  topic: string;
  researchContext: OpenAICompatibleContext
}
```

EvanBoyle (commented):
I think we should give the context type a name (left a more specific comment about this). This will enable a pattern for better composability. We should also create an example that shows using that pattern. 

There is still one thing that feels unsolved across all of this: model providers don't share model names. It isn't enough to just pass a provider + completion component. There still needs to be a decision somewhere on what what model to use. I wonder if the `OpenAICompatibleContext` should include a `defaultModel` param? I think that would smooth things over a lot. Let me know what you think. 

For the example, I think the best one would be to create a `WebResearchComponent` 

This component (1) accept a general purpose context for summarizing research, and (2) optionally accept a specialized research context. Then you could do things like: 

1. Use openai for both research and synthesis
2. Use perplexity for research and openai for synthesis
3. use perplexity for research and anthropic for synthesis

I think having an example that does this would cover all of the bases we need here. Curious for thoughts from @dereklegenzoff as well.

jmoseley:
> There still needs to be a decision somewhere on what what model to use. I wonder if the OpenAICompatibleContext should include a defaultModel param? I think that would smooth things over a lot. Let me know what you think.

I did account for models (you can specify a `ModelName` type parameter to constrain the values, otherwise it is just a `string`. We could roll it into the output from `createOpenAIClientContext` so that it does not need to be a separate parameter for the parameterized version, but that sort of strongly binds your context to a single model, which is not always the case.

EvanBoyle:
> I did account for models (you can specify a `ModelName` type parameter to constrain the values, otherwise it is just a `string`.

Not clear to me that this actually helps us accomplish anything. Those types aren't available at runtime for the reusable component author to figure out what models are valid to select. Unless I'm missing something.

dereklegenzoff:
I don't have any particularly well formed thoughts but the scenario jumping out at me is:
- In a workflow you want to use small and large models from the same provider (e.g. gpt-4o and gpt-4o-mini). There's also a three tier version of this (haiku, sonnet, opus or llama 8b, 70b, 405b)
- As a best practice, customers may want to pin to a particular version for each (e.g. gpt-4o-2024-08-06) and use that as the default small/large model
- Often you'll upgrade everything at once but you may keep some components on an older version if results are worse with the newer one

I don't think a `defaultModel` param really solves for this that well but at the same time you could accomplish the scenario above well enough in the current implementation.

jmoseley:
Here is what I was thinking with this implementation, but I had an idea last night, and it is below:

```
export const WebResearchAndSummarize = gsx.StreamComponent<{
  prompt: string;
  summarizeContext: Context<OpenAI>;
  webSearchContext: Context<OpenAI>;
  summarizeModel: string;
  webSearchModel: string;
}>(
  "ResearchAndSummarize",
  async function ({
    prompt,
    summarizeContext: openAiContext,
    webSearchContext: perplexityContext,
    summarizeModel,
    webSearchModel,
  }) {
    const webResearch = await gsx.execute<string>(
      <ChatCompletion
        context={perplexityContext}
        model={webSearchModel}
        messages={[
          {
            role: "system",
            content:
              "You are a helpful assistant that takes a prompt from the user and returns raw information from the web. Only return the raw information related to the prompt, no other text.",
          },
          {
            role: "user",
            content: prompt,
          },
        ]}
      />,
    );

    return (
      <ChatCompletion
        context={openAiContext}
        model={summarizeModel}
        messages={[
          {
            role: "system",
            content:
              "You are an export research assistant. You will be given a prompt/topic and some broad web research, and your task is to extract the most relevant information and return it in a concise manner.",
          },
          {
            role: "user",
            content: `${prompt}\n\n<web_research>${webResearch}</web_research>`,
          },
        ]}
      />
    );
  },
);
```

However, I had a brain wave laying in bed at 4am last night (thanks Charlie! 👶 ). With a tweak to the shape for `createOpenAIClientContext`, I don't think we need parameterized contexts. Passing the models is still necessary, but providers remove the need for the context params above, and this is more familiar to a React developer (read the comments for more details and options):

```
// We could support a model param as part of the context, or allow the developer to extend the context however they want.
// The developer could also choose to use any old context, and implement it however they need.
const { Provider: SummarizeClientProvider, ChatCompletion: SummarizeChatCompletion } = createOpenAIClientContext<{ model: string }>();
const { Provider: WebSearchClientProvider, ChatCompletion: WebSearchChatCompletion } = createOpenAIClientContext();

export { SummarizeClientProvider, WebSearchClientProvider };

export const WebResearchAndSummarize = gsx.StreamComponent<{
  prompt: string;
  summarizeModel: string;
  webSearchModel: string;
}>(
  "ResearchAndSummarize",
  async function ({
    prompt,
    summarizeModel,
    webSearchModel,
  }) {
    const webResearch = await gsx.execute<string>(
      <QueryChatCompletion
        model={webSearchModel}
        messages={[
          {
            role: "system",
            content:
              "You are a helpful assistant that takes a prompt from the user and returns raw information from the web. Only return the raw information related to the prompt, no other text.",
          },
          {
            role: "user",
            content: prompt,
          },
        ]}
      />,
    );

    return (
      <SummarizeChatCompletion
        model={summarizeModel}
        messages={[
          {
            role: "system",
            content:
              "You are an export research assistant. You will be given a prompt/topic and some broad web research, and your task is to extract the most relevant information and return it in a concise manner.",
          },
          {
            role: "user",
            content: `${prompt}\n\n<web_research>${webResearch}</web_research>`,
          },
        ]}
      />
    );
  },
);

// To use it:

import { SummarizeClientProvider, WebSearchClientProvider, WebResearchAndSummarize } from 'custom-gensx-component';

await gsx.execute<string>(
  <SummarizeClientProvider client={new OpenAI({ apiKey: '...' })} model="gpt-40"> // model here could be optional or whatever, upto the developer of the shared component
    <WebSearchClientProvider client={new OpenAI({ apiKey: '...', baseUrl: 'perplexity.ai' })>
      <WebResearchAndSummarize prompt="LLMs strengths and weaknesses" summarizeModel="gpt-4o" webSearchModel="perpelxity-model" />
    </WebSearchClientProvider>
  </SummarizeClientProvider>
);
```

This is much more React-esque, and removes the necessity to pass context as a parameter. It _could_ solve the model problem as well, since the developer of the shared component could choose to include model in the context instead of parameterizing the component, or even both.

The one thing that will be really challenging with this model (either implementation) is strongly typing the model name, since that is wholly dependent on the `client` param, so it would need to be a type parameter for the components (that cannot be inferred), and JSX does not support that. But the world has survived just fine without that so far.

EvanBoyle:
@jmoseley is the key insight here that a reusable component declares it's own providers that you need to implement yourself? I think that model makes sense. 

I would be curious for @dereklegenzoff's feedback on this as well.

dereklegenzoff:
in that last option, would a user always need to pass in the client even when using the provider with openai?

```tsx
<OpenAIProvider client={new OpenAI({ apiKey: '...',})>
```

Or could they also do

```tsx
<OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>

<WebSearchClientProvider apiKey={process.env.OPENAI_API_KEY} baseUrl={ 'perplexity.ai' })>
```

I think it'd be weird if you always have to pass a client but otherwise i like it.

danfhernandez:
To be honest, this is starting to feel more complicated then just using the sdk and/or raw ts... I worry that the provider model is overkill for what it's trying to solve. I.e. a chat completion could simply be parameterized and we call it a day. You can maybe set yourself up a component that wraps it with the api key included or something, idk and just use that. But perhaps I'm missing something critical here. 

That said, it feels like having to wrap a component in a context feels less natural and confusing compared to parameterizing it. But if we can do both, perhaps it'll be useful to have to see what usage patterns form with users that may be familiar with this type of pattern.

EvanBoyle:
@danfhernandez definitely hear the feedback there. We don't want to create something needlessly complicated. 

But, even if we were to just use the openAI SDK within the component the problem we're trying to solve doesn't go away. To truly build a reusable component, we have to have a model so that users can use any given published component with their model provider and model of choice. I don't think we can sustain an ecosystem when "this component only works for OpenAI" and "this other one only works with anthropic models", feels unsustainable. 

Do you have ideas on how we might solve that problem? 

I think I agree with you that the version that passes these contexts/providers as props is pretty simple and easy to understand. I read that code and instantly grok'd what it does. The second example took a few minutes of reading to understand.

jmoseley:
>is the key insight here that a reusable component declares it's own providers that you need to implement yourself? I think that model makes sense.

Yes that's right. This is how any reasonably complex React thing works (eg auth). A service provider might provide a bunch of hooks or components, and they expect you to drop a provider somewhere in your component tree to power them.

This model might get a bit tedious, because your workflow might have a bunch of different Providers for various components that are all effectively providing an OpenAI client, but I don't see any other mechanism that will be intuitive and flexible. We can assume that these shared components might be complex enough to need contexts that have parameters beyond just the OpenAI client. 

@danfhernandez I hear your feedback, and I think I largely agree with you for the small scale. But if we expand the scope of these things, I think it makes more sense. For example, imagine we wanted to share our current article creation workflow as a component that someone can use, and make it pluggable so they can drop in whichever LLMs they wanted. The providers would enable us to parameterize the models without needing to prop drill, giving flexibility to the caller.

It seems like the alternative is to introduce a Langchain-esque `Runnable`, but even that needs to be configured and injected into the component somewhere, and Provider feels cleaner than prop drilling.

EvanBoyle:
@danfhernandez I agree this is an "advanced topic", and I don't expect users to write their local components this way. But I'm nearly certain we will want to publish examples that let users pick from claude, openAI, gemeni, etc, and optionally inject perplexity or other web models for certain steps. I expect that people who want to publish components for broad usage and adoption on NPM will need to find patterns for doing the same thing. 

If there are alternative patterns that are simpler, would be great to see an example. But - I feel like finding a pattern where "everything is reusable as long as the API is openAI compatible" is a very compelling starting point. 

Upon further examination of these two patterns, I feel like option 2 is more verbose and less composable. You have to instantiate two providers and a component when you want to use the single component in the example. 

For option 1, you might already have a context laying around that you can pass in and reuse. That seems much better IMO, and composition via parameters is much easier to read and grok. Option (2) feels more like rocket science. Even if this is how you do things like Auth in react, I doubt that many react engineers are regularly touching patterns like this or implementing auth on their own - small percentage of people doing this work.

EvanBoyle:
Closing this out in favor of the pattern described here: https://github.com/gensx-inc/gensx/issues/74#issuecomment-2625394687

Issue #136, opened 2025-01-18
[Epic] Checkpoint implementation


EvanBoyle:
We need to create a checkpoint implementation that serializes the component tree for visualization, tracing, and debugging as a workflow executes. We'll start with a basic implementation that posts the entire JSON object, and iterate towards a journaling API with a smaller payload if necessary. Original prototype branch here: https://github.com/cortexclick/gensx/compare/evan/checkpoints

## Task list:
- [x] #137
- [x] #138
- [x] #139
- [x] #140
- [x] #141
- [x] #142
- [x] #144

PR #143, opened 2025-01-19
feat: Add checkpointing

### LINKED ISSUE #137: [Checkpoints] Implement naive checkpointing
Naive implementation that just writes a JSON file (or calls an API)

### LINKED ISSUE #138: [Checkpoints] Differentiate between children and nested children
The checkpoint implementation only has a single `children` element on each node. In reality we need to differentiate between: 

1. children created within the component function
2. children created via nesting

### LINKED ISSUE #139: [Checkpoints] Differentiate output and final result from nested children
Any given component produces an output (it's return) and has a final value that is produced via the nested children chain. We should record both of these and differentiate between them in the checkpoint.

### LINKED ISSUE #140: [Checkpoints] Ensure that start time and end time are correct

### LINKED ISSUE #141: [Checkpoints] Optimize code path to ensure checkpoint writing is non-blocking and queued
The naive implementation blocks workflow execution on writing a checkpoint every time a node is added or completes. Workflow execution should not block, and we should avoid doing checkpoint writes in rapid succession where possible. 

We can both make this async (checkpoint in the background) and use some sort of queueing mechanism such that we only write the most recent checkpoint in the queue at any given point in time and throw away the rest.

### LINKED ISSUE #144: [Checkpoints] Support calling `gensx-console` API

EvanBoyle:
Checkpoint implementation with tests. 

6/7 items for https://github.com/cortexclick/gensx/issues/136 - just need to add support for streaming components (will require some care to fork the stream and make sure we don't block - will do this separately). 

Important bits here: 
- checkpointing is hidden behind an env var `GENSX_CHECKPOINTS` and is off by default
- Care is taken to minimize chattiness with the server via a "queue". We only ever process one checkpoint at a time. Processing checkpoints is non-blocking. We drop "intermediate" checkpoints if we have a "fresh" checkpoint in the queue. There is not really a queue, nor a concept of "fresh checkpoints" - but this is the best analogy. 
- Parent ID is maintained throughout execution via context to ensure nodes are parented properly. This includes subworkflows created via gsx.execute within components. 

Fixes #137 fixes #138 fixes #139 fixes #140 fixes #141 fixes #144

### DIFF
--- a/packages/gensx/package.json
+++ b/packages/gensx/package.json
@@ -29,7 +29,7 @@
     "build": "vite build",
     "dev": "vite build --watch",
     "clean": "rm -rf dist",
-    "test": "vitest run checkpoint --coverage",
+    "test": "vitest run --coverage",
     "test:watch": "vitest watch",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix"
--- a/packages/gensx/package.json
+++ b/packages/gensx/package.json
@@ -29,7 +29,7 @@
     "build": "vite build",
     "dev": "vite build --watch",
     "clean": "rm -rf dist",
-    "test": "vitest run --coverage",
+    "test": "vitest run checkpoint --coverage",
     "test:watch": "vitest watch",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix"
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -38,9 +38,8 @@ export interface ExecutionNode {
 }
 
 export interface CheckpointWriter {
-  currentNode?: ExecutionNode;
   root?: ExecutionNode;
-  addNode: (node: Partial<ExecutionNode>) => Promise<string>;
+  addNode: (node: Partial<ExecutionNode>, parentId?: string) => Promise<string>;
   completeNode: (id: string, output: unknown) => void;
   addMetadata: (id: string, metadata: Record<string, unknown>) => void;
   write: () => void;
@@ -50,7 +49,6 @@ export interface CheckpointWriter {
 export class CheckpointManager implements CheckpointWriter {
   private nodes = new Map<string, ExecutionNode>();
   public root?: ExecutionNode;
-  public currentNode?: ExecutionNode;
   public checkpointsEnabled: boolean;
 
   // Track active checkpoint write
@@ -121,7 +119,7 @@ export class CheckpointManager implements CheckpointWriter {
     }
   }
 
-  async addNode(partial: Partial<ExecutionNode>) {
+  async addNode(partial: Partial<ExecutionNode>, parentId?: string) {
     const node: ExecutionNode = {
       id: await generateUUID(),
       componentName: "Unknown",
@@ -133,16 +131,16 @@ export class CheckpointManager implements CheckpointWriter {
 
     this.nodes.set(node.id, node);
 
-    if (!this.root) {
-      // First node becomes the root
+    if (parentId) {
+      const parent = this.nodes.get(parentId);
+      if (parent) {
+        node.parentId = parentId;
+        parent.children.push(node);
+      }
+    } else if (!this.root) {
       this.root = node;
-    } else if (this.currentNode) {
-      // Add as child to current node
-      node.parentId = this.currentNode.id;
-      this.currentNode.children.push(node);
     }
 
-    this.currentNode = node;
     this.updateCheckpoint();
     return node.id;
   }
@@ -152,14 +150,6 @@ export class CheckpointManager implements CheckpointWriter {
     if (node) {
       node.endTime = Date.now();
       node.output = output;
-
-      if (node.parentId) {
-        const parent = this.nodes.get(node.parentId);
-        this.currentNode = parent;
-      } else {
-        this.currentNode = undefined;
-      }
-
       this.updateCheckpoint();
     } else {
       console.warn(`[Tracker] Attempted to complete unknown node:`, { id });
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -1,0 +1,152 @@
+// Platform-agnostic file operations
+async function writeFileSafe(path: string, data: string): Promise<void> {
+  const isNode = typeof process !== "undefined";
+
+  if (isNode) {
+    // Node.js environment
+    try {
+      const { writeFile } = await import("fs/promises");
+      await writeFile(path, data);
+    } catch (error) {
+      console.error(`[Tracker] Failed to write file:`, { path, error });
+      throw error;
+    }
+  } else {
+    // Browser environment - could implement browser-specific storage here
+    console.warn(
+      "[Tracker] File writing is not supported in browser environment",
+    );
+  }
+}
+
+// Cross-platform UUID generation
+async function generateUUID(): Promise<string> {
+  try {
+    // Try Node.js crypto first
+    const crypto = await import("node:crypto");
+    return crypto.randomUUID();
+  } catch {
+    // Fallback to browser crypto
+    if (typeof globalThis !== "undefined") {
+      return globalThis.crypto.randomUUID();
+    }
+    // Simple fallback for environments without crypto
+    return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, c => {
+      const r = (Math.random() * 16) | 0;
+      const v = c === "x" ? r : (r & 0x3) | 0x8;
+      return v.toString(16);
+    });
+  }
+}
+
+export interface ExecutionNode {
+  id: string;
+  componentName: string;
+  parentId?: string;
+  startTime: number;
+  endTime?: number;
+  props: Record<string, unknown>;
+  output?: unknown;
+  children: ExecutionNode[];
+  metadata?: {
+    logs?: string[];
+    tokenCounts?: {
+      input: number;
+      output: number;
+    };
+    [key: string]: unknown;
+  };
+}
+
+export interface CheckpointWriter {
+  currentNode?: ExecutionNode;
+  root: ExecutionNode;
+  addNode: (node: Partial<ExecutionNode>) => Promise<string>;
+  completeNode: (id: string, output: unknown) => Promise<void>;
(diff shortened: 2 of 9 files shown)


packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.updateCheckpoint();
+      }
+    });
+  }
+
+  private async writeCheckpoint() {
+    if (!this.root) return;
+
+    try {
+      const response = await fetch("http://localhost:3000/api/execution", {
```
jmoseley, line 103: Any reason not to make this configurable right away?

```suggestion
      const baseUrl = process.env.GENSX_CHECKPOINT_URL || 'https://localhost:3000';
      const url = path.join(baseUrl, 'api/execution');
      const response = await fetch(url, {
```

packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,180 @@
...
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(this.root),
```
jmoseley, line 108: We write the entire tree, every time?

packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,180 @@
...
+    const checkpointsEnv = process.env.GENSX_CHECKPOINTS?.toLowerCase();
+    this.checkpointsEnabled =
+      checkpointsEnv === "true" ||
+      checkpointsEnv === "1" ||
+      checkpointsEnv === "yes";
+  }
+
+  // updateCheckpoint POSTs the current component tree to the GenSX API.
+  // Special care is taken to do this in a non-blocking manner, and in a way that
+  // minimizes chattiness with the server.
```
jmoseley, line 70: Wouldn't a buffer model work better to reduce chattiness?

```
while(true) {
   await setTimeout(1000);
   if (updateHappened) {
      this.writeCheckpoint();
   }
}
```

You need a flush step at the end, but I think we already need that (for the case where the workflow ends while the second last checkpoint is being written)

packages/gensx/package.json
```diff
@@ -29,7 +29,7 @@
     "build": "vite build",
     "dev": "vite build --watch",
     "clean": "rm -rf dist",
-    "test": "vitest run --coverage",
+    "test": "vitest run checkpoint --coverage",
```
jmoseley, line 32: Whats this?

packages/gensx/src/checkpoint.ts
EvanBoyle: Yup. We can replace this with a journaling API in the future if we find performance to be an issue. However, all of these calls are non-blocking and we make the minimal number of requests necessary.

packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,170 @@
...
+  endTime?: number;
+  props: Record<string, unknown>;
+  output?: unknown;
+  children: ExecutionNode[];
+  metadata?: {
+    logs?: string[];
+    tokenCounts?: {
+      input: number;
+      output: number;
+    };
```
jmoseley, line 35: These aren't super meaningful without also knowing the provider/model, so you can calculate cost.

```suggestion
    tokenCounts?: {
      input: number;
      output: number;
      provider: string;
      model: string
    };
```

packages/gensx/src/checkpoint.ts
EvanBoyle: The top level goals I had were: 

1. make sure that writing checkpoints doesn't block the workflow from executing
2. ensure the console can visualize workflows "responsively"
3. make sure that we don't do more work than is necessary -- ie write at most one checkpoint at any given point in time, don't repeat intermediate updates if we have a never checkpoint available

The current implementation satisfies all of these. We could loop and sleep, but not clear that this is strictly necessary right now. Maybe we find out otherwise under load, but I'd prefer to keep things easier to reason about for the time being.

packages/gensx/package.json
EvanBoyle: leftover from development, will remove.

packages/gensx/tests/checkpoint.test.tsx
```diff
@@ -0,0 +1,359 @@
...
+      a: "a:first",
+      b: "b:second",
+    });
+
+    // Get final checkpoint state
+    const finalCheckpoint = checkpoints[checkpoints.length - 1];
+
+    // Verify checkpoint structure
+    expect(finalCheckpoint).toMatchObject({
+      componentName: "ParentComponent",
```
jmoseley, line 279: Do you think we are losing valuable information without including the output shape?

An alternative result here could be:

```
{
  comonentName: "ParentComponent",
  output: {
    a: {
      componentName: "ComponentA",
      ...
    },
    b: {
      componentName: "ComponentB",
      ...
    }
  },
}
```

packages/gensx/src/checkpoint.ts
EvanBoyle: We aren't using these fields right now. I expect we'll update the data model when we do add support for token tracking. Will leave until that point if that is ok with you - otherwise I can just remove this stuff for now.

packages/gensx/src/resolve.ts
```diff
@@ -53,5 +53,8 @@ export async function resolveDeep<T>(value: unknown): Promise<T> {
  * This is the main entry point for executing workflow components.
  */
 export async function execute<T>(element: ExecutableValue): Promise<T> {
-  return (await resolveDeep(element)) as T;
+  const context = getCurrentContext().getWorkflowContext();
+  const result = (await resolveDeep(element)) as T;
+  context.checkpointManager.write();
```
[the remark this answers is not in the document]
EvanBoyle: My understanding is that the node process will continue running until all active promises are resolved, no? Awaiting it would cause all calls to execute to have a blocking call as well.

packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,180 @@
+// Cross-platform UUID generation
+async function generateUUID(): Promise<string> {
+  try {
+    // Try Node.js crypto first
+    const crypto = await import("node:crypto");
+    return crypto.randomUUID();
+  } catch {
+    // Fallback to browser crypto
+    if (typeof globalThis !== "undefined") {
+      return globalThis.crypto.randomUUID();
```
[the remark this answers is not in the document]
jmoseley: Apparently not. https://dev.to/simplr_sh/ditch-the-import-why-cryptorandomuuid-is-your-new-best-friend-for-uuids-2lp3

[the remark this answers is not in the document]
EvanBoyle: Yeah, the only async part is the import - which I believe we only pay for the impact of once. I have no data on how fast the call to generate a uuid is, but I don't think there is a way around generating a unique identifier per node in the graph. If we find this is a bottle neck in the future we can look at optimizing.

packages/gensx/src/checkpoint.ts
jmoseley: (2) is a good requirement that a buffer model would not do well, and more server calls for that requirement seems worth it.

packages/gensx/src/checkpoint.ts
```diff
@@ -0,0 +1,170 @@
...
+          status: response.status,
+          message: await response.text(),
+        });
+      }
+    } catch (error) {
+      console.error(`[Checkpoint] Failed to save checkpoint:`, { error });
+    }
+  }
+
+  async addNode(partial: Partial<ExecutionNode>, parentId?: string) {
```
jmoseley, line 122:
```suggestion
  async addNode(partialNode: Partial<ExecutionNode>, parentId?: string) {
```

packages/gensx/src/resolve.ts
jmoseley, line 58: If you don't await this, we might lose it if the program immediately ends right? I think this should probably be an awaitable `flush`.

packages/gensx/tests/checkpoint.test.tsx
```diff
@@ -0,0 +1,359 @@
...
+    const fetchCalls = (global.fetch as unknown as ReturnType<typeof vi.fn>)
+      .mock.calls.length;
+
+    // We expect:
+    // - Some minimum number of calls to capture the state (could be heavily batched)
+    // - Less than the theoretical maximum (303 = parent(2) + children(200) + execute calls(101))
+    // - Evidence of queueing (significantly less than theoretical maximum)
+    expect(fetchCalls).toBeGreaterThan(0); // At least some calls must happen
+    expect(fetchCalls).toBeLessThan(303); // Less than theoretical maximum
+    expect(fetchCalls).toBeLessThan(250); // Evidence of significant queueing
```
jmoseley, line 175: Any concerns about this being flakey?

packages/gensx/tests/checkpoint.test.tsx
EvanBoyle: updated.

packages/gensx/src/checkpoint.ts
EvanBoyle: Done.

jmoseley (approved):

packages/gensx/tests/checkpoint.test.tsx
EvanBoyle: 303 is the max number of calls it can theoretically make. 

everytime I've run it it has been exactly 204 (or maybe 205, I can't remember) - but if it turns out to be flakey we can remove the middle constraint.

packages/gensx/src/checkpoint.ts
EvanBoyle: done

packages/gensx/src/resolve.ts
jmoseley: In theory. Not if there is a `process.exit` kicking around, or if its in a serverless environment.

[Langsmith](https://docs.smith.langchain.com/observability/how_to_guides/tracing/serverless_environments) provides an env var for this, [Sentry](https://docs.sentry.io/platforms/javascript/guides/aws-lambda/configuration/draining/) suggests you need to manually call `.close`.

packages/gensx/src/checkpoint.ts
jmoseley: Perhaps removing would be the right choice then 🤷

PR #146, opened 2025-01-20
docs: adding example page for hn analyzer and blog writer

dereklegenzoff:
Adds a basic examples doc page for the hn analyzer example

### DIFF
--- a/docs/src/content/docs/examples/blog-writing.mdx
+++ b/docs/src/content/docs/examples/blog-writing.mdx
@@ -78,7 +78,7 @@ const LLMEditor = gsx.StreamComponent<LLMEditorProps>(
 );
 ```
 
-Then when the component is invoked, `stream={true}` is passed to the component so that the output is streamed back and can be surfaced to the user.
+Then when the component is invoked, `stream={true}` is passed to the component so that the output is streamed back and can be surfaced to the user:
 
 ```tsx
 <LLMEditor draft={draft} stream={true} />
--- a/docs/src/content/docs/examples/blog-writing.mdx
+++ b/docs/src/content/docs/examples/blog-writing.mdx
@@ -5,10 +5,85 @@ sidebar:
   order: 2
 ---
 
-GenSX is a simple framework for building complex generative AI and LLM workflows - think agents, chatbot APIs, RAG, reflection, and more.
+The [Blog Writer example](https://github.com/cortexclick/gensx/tree/main/examples/blogWriter) demonstrates how to create a multi-step blog writing workflow using GenSX. It combines research, content generation, and editing to create a polished blog post based on a prompt.
 
-It uses `jsx` to define and orchestrate workflows with functional, reusable components:
+## Workflow
+
+The Blog Writer workflow consists of the following steps:
+
+1. Parallel research phase:
+   - Brainstorm topics using LLM (`<LLMResearchBrainstorm>`)
+   - Research each topic in detail (`<LLMResearch>`)
+   - Gather web research (`<WebResearcher>`)
+2. Write initial draft based on research (`<LLMWriter>`)
+3. Edit and polish the content (`<LLMEditor>`)
+
+## Running the example
+
+```bash
+# Install dependencies
+pnpm install
+
+# Set your OpenAI API key
+export OPENAI_API_KEY=<your_api_key>
+
+# Run the example
+pnpm run run
+```
+
+## Key patterns
+
+### Running components in parallel
+
+The `<ParallelResearch>` component runs both the `LLMResearchBrainstorm` and `WebResearcher` components in parallel. Both sub-workflows return an array of strings which are automatically combined into a single array.
 
 ```tsx
-
+const ParallelResearch = gsx.Component<
+  ParallelResearchComponentProps,
+  ParallelResearchOutput
+>("ParallelResearch", ({ prompt }) => {
+  return (
+    <>
+      <LLMResearchBrainstorm prompt={prompt}>
+        {({ topics }) => {
+          return topics.map((topic) => <LLMResearch topic={topic} />);
+        }}
+      </LLMResearchBrainstorm>
+      <WebResearcher prompt={prompt} />
+    </>
+  );
+});
 ```
+
+### Streaming Output
+
+The workflow also shows how to stream back the final output with GenSX. The `<LLMEditor>` component is a `StreamComponent` and the `<ChatCompletion>` component has `stream={true}`.
+
+```tsx
+const LLMEditor = gsx.StreamComponent<LLMEditorProps>(
+  "LLMEditor",
+  ({ draft }) => {
+    return (
+      <ChatCompletion
+        stream={true}
+        model="gpt-4o-mini"
+        temperature={0}
+        messages={[
+          { role: "system", content: systemPrompt },
+          { role: "user", content: draft },
+        ]}
+      />
+    );
+  },
+);
+```
+
+Then when the component is invoked, `stream={true}` is passed to the component so that the output is streamed back and can be surfaced to the user.
+
+```tsx
+<LLMEditor draft={draft} stream={true} />
+```
+
+## Additional resources
+
+Check out the other GenSX examples in the [Github Repo](https://github.com/cortexclick/gensx).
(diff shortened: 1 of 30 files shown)


docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
 
-GenSX is a simple framework for building complex generative AI and LLM workflows.
+The [Hacker News Analyzer](https://github.com/cortexclick/gensx/tree/main/examples/hackerNewsAnalyzer) example uses GenSX to fetch, analyze, and extract trends from the top Hacker News posts. It shows how to combine data fetching, parallel analysis, and content generation in a single workflow to create two outputs: a detailed report and a tweet of the trends.
 
-It uses `jsx` to define and orchestrate workflows with functional, reusable components:
+## Workflow
+
+The Hacker News Analyzer workflow is composed of the following steps:
+
+1. Fetch the top 500 Hacker News posts (`<HnCollector>`)
```
EvanBoyle, line 14: Might be worth noting that this is limited to `text` posts. We skip analyzing any link posts that would require web scraping for convenience of the example.

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+2. Process each post in parallel (`<AnalyzeHNPosts>`)
+   a. Summarize the content (`<PostSummarizer>`)
+   b. Analyze the comments (`<CommentsAnalyzer>`)
```
EvanBoyle, line 17: Formatting for these is off in the preview: 
image

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+3. Writes a detailed report identifying the key trends across all posts (`<TrendsAnalyzer>`)
+4. Edits the report into the style of Paul Graham (`<PGEditor>`)
+5. Generates a tweet in the voice of Paul Graham (`<PGTweetWriter>`)
```
EvanBoyle, line 20: Both of these steps in the workflow don't do a particularly good job at emulating paul graham. Might be worth following up with a tweak to those two examples that uses few shot prompting whenever we have time.

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+
+```bash
+# Install dependencies
+pnpm install
+
+# Set your OpenAI API key
+export OPENAI_API_KEY=<your_api_key>
+
+# Run the example
+pnpm run run
```
EvanBoyle, line 32: I've always felt weird about typing out `run run` wondering if we should come up with a better standard for our examples now that we have to put this stuff in writing lol.

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+- `hn_analysis_report.md`: A detailed analysis report
+- `hn_analysis_tweet.txt`: A tweet-sized summary of the analysis
+
+## Key patterns
+
+### Parallel processing
+
+The workflow uses the `<AnalyzeHNPosts>` component to process each post in parallel. To do this, it takes advantage of the ability to have components return arrays and structured objects containing other components.
+
+This component takes a list of stories and returns a list of `analyses` with each containing a summary and a comment analysis for the post.
```
EvanBoyle, line 46: Some additional things we want to highlight clearly here: 

1. You can use plain old `array.map` to create lists of elements
2. JSX support is extended to objects, so any part of your return object can have a component that will get resolved at runtime.

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+        ),
+      })),
+    };
+  },
+);
+```
+
+### Multiple outputs
 
+This example also shows a simple pattern returning multiple outputs from a workflow. In this case, both the outputs of `<TrendAnalyzer>` and `<PGTweetWriter>` are returned as a single object. This is done by leveraging the the child function of `<PGEditor>` to combine the outputs of multiple components into a single object.
```
EvanBoyle, line 66: Let's use something other than leveraging. Maybe "using"?

docs/src/content/docs/examples/hn.mdx
```diff
@@ -1,14 +1,87 @@
...
+        </PGTweetWriter>
+      )}
+    </PGEditor>
+  )}
+</TrendAnalyzer>
 ```
+
+## Additional resources
+
+Check out the other GenSX examples in the [Github Repo](https://github.com/cortexclick/gensx).
```
EvanBoyle, line 87: Maybe link to the examples folder/readme?

docs/src/content/docs/examples/blog-writing.mdx
```diff
@@ -1,14 +1,89 @@
 ---
-title: Blog writing
+title: Blog Writer
 description: Blog writing agent with Perplexity and GenSX
 sidebar:
   order: 2
 ---
 
-GenSX is a simple framework for building complex generative AI and LLM workflows - think agents, chatbot APIs, RAG, reflection, and more.
+The [Blog Writer example](https://github.com/cortexclick/gensx/tree/main/examples/blogWriter) demonstrates how to create a multi-step blog writing workflow using GenSX. It combines research, content generation, and editing to create a polished blog post based on a prompt.
```
EvanBoyle, line 8: Maybe we should add a little color here, something like: 

Breaking down complex jobs into smaller, discrete tasks is one of the best ways to improve the quality of LLM outputs. The blog writing workflow does this by mimicking the way that a human would write a blog post: first conducting research, then generating a first draft, and finally editing that draft.

docs/src/content/docs/examples/blog-writing.mdx
```diff
@@ -1,14 +1,89 @@
...
+  ParallelResearchOutput
+>("ParallelResearch", ({ prompt }) => {
+  return (
+    <>
+      <LLMResearchBrainstorm prompt={prompt}>
+        {({ topics }) => {
+          return topics.map((topic) => <LLMResearch topic={topic} />);
+        }}
+      </LLMResearchBrainstorm>
+      <WebResearcher prompt={prompt} />
```
EvanBoyle, line 52: Fine to merge this, but let's not forget to come back and actually implement the web research component. When we do so, we'll have to update the prereqs section of this doc to include perplexity or whatever else we choose.

docs/src/content/docs/examples/blog-writing.mdx
```diff
@@ -1,14 +1,89 @@
...
+    </>
+  );
+});
+```
+
+### Streaming Output
+
+The workflow also shows how to stream back the final output with GenSX. The `<LLMEditor>` component is a `StreamComponent` and the `<ChatCompletion>` component has `stream={true}`.
```
EvanBoyle, line 60: Maybe change the voice a little? Something like "This workflow streams the final output to reduce the time that it takes for the user to receive the first token. 

Also, should take this as an opportunity to cross link to the StreamComponent docs.

EvanBoyle (approved):
A few comments, but overall looks good!

docs/src/content/docs/examples/hn.mdx
dereklegenzoff: oh yeah 100% agree - run run is terrible. I'll submit a PR today to improve this on all of the examples.

docs/src/content/docs/examples/hn.mdx
dereklegenzoff: great point; will update

docs/src/content/docs/examples/hn.mdx
dereklegenzoff: good catch

docs/src/content/docs/examples/hn.mdx
dereklegenzoff: sounds good, i agree it makes sense to rev on these before we launch.

docs/src/content/docs/examples/blog-writing.mdx
dereklegenzoff: agreed, noted. Makes me think we'll need to add a prereq section once we require more than one key.

docs/src/content/docs/examples/blog-writing.mdx
dereklegenzoff: i like it, updated and added a link

docs/src/content/docs/examples/hn.mdx
dereklegenzoff: updated to better highlight these

PR #148, opened 2025-01-20
docs: adding provider docs

dereklegenzoff:
Adds the concept doc for providers

Also adds a provider example that shows how to create a provider for Firecrawl (that is discussed in the doc)

It's worth noting that the top of this doc is the same as the text from the basic concepts page but it extends the info. Eventually we will also want an "Available Providers" section but when written currently, it only contains the OpenAIProvider.

### DIFF
--- a/docs/src/content/docs/basic-concepts.mdx
+++ b/docs/src/content/docs/basic-concepts.mdx
@@ -203,3 +203,5 @@ Providers offer several benefits:
 - Provide a simpler, more focused API than raw contexts
 - Keep configuration separate from business logic
 - Enable dependency injection for testing and development
+
+For more details on providers, see the [Providers](../concepts/providers) page.
--- a/docs/src/content/docs/concepts/providers.mdx
+++ b/docs/src/content/docs/concepts/providers.mdx
@@ -1,14 +1,96 @@
 ---
 title: Providers
-description: GenSX architecture and design goals
+description: Learn how to manage API keys and configuration dependencies in GenSX using providers
 sidebar:
   order: 3
 ---
 
-GenSX is a simple framework for building complex generative AI and LLM workflows.
+Providers are a specialized use of [contexts](../contexts) that focus on managing configuration and dependencies for your workflow. They're built using the context system described above, but provide a more focused API for configuration management.
 
-It uses `jsx` to define and orchestrate workflows with functional, reusable components:
+The main provider available today is the `OpenAIProvider`, which manages your OpenAI API key:
 
 ```tsx
+const result = await gsx.execute(
+  <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
+    <ChatCompletion
+      model="gpt-4o-mini"
+      messages={[{ role: "user", content: "Hello!" }]}
+    />
+  </OpenAIProvider>,
+);
+```
 
+Providers offer several benefits:
+
+- Manage complex configuration like API clients and settings
+- Provide a simpler, more focused API than raw contexts
+- Keep configuration separate from business logic
+- Enable dependency injection for testing and development
+
+## Creating a Provider
+
+If you want to use a provide that isn't available out of the box, you can easily create your own. The example below shows how to create a provider for the [Firecrawl](https://www.firecrawl.dev/) API.
+
+Start by importing `gsx` and the package you want to use:
+
+```tsx
+import { gsx } from "gensx";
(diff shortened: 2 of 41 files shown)


docs/src/content/docs/basic-concepts.mdx
```diff
@@ -203,3 +203,5 @@ Providers offer several benefits:
 - Provide a simpler, more focused API than raw contexts
 - Keep configuration separate from business logic
 - Enable dependency injection for testing and development
+
+For more details on providers, see the [Providers](../concepts/providers) page.
```
EvanBoyle, line 207: Without a preview I'm not 100% certain but I think this should be `../../concepts/providers`

docs/src/content/docs/concepts/providers.mdx
```diff
@@ -1,14 +1,96 @@
 ---
 title: Providers
-description: GenSX architecture and design goals
+description: Learn how to manage API keys and configuration dependencies in GenSX using providers
 sidebar:
   order: 3
 ---
 
-GenSX is a simple framework for building complex generative AI and LLM workflows.
+Providers are a specialized use of [contexts](../contexts) that focus on managing configuration and dependencies for your workflow. They're built using the context system described above, but provide a more focused API for configuration management.
```
EvanBoyle, line 8: somewhere in the doc we should mention that this is a solution for "prop drilling".

examples/providers/README.md
```diff
@@ -0,0 +1,20 @@
+# Providers Example
```
EvanBoyle, line 1: Is this really a provider example? Or is it a web scraping/firecrawl example that happens to show you how to create a provider? Should we just call this "Web scraping example"?

docs/src/content/docs/concepts/providers.mdx
```diff
@@ -1,14 +1,96 @@
...
+  </OpenAIProvider>,
+);
+```
+
+Providers offer several benefits:
+
+- Manage complex configuration like API clients and settings
+- Provide a simpler, more focused API than raw contexts
+- Keep configuration separate from business logic
+- Enable dependency injection for testing and development
```
EvanBoyle, line 28: I feel like this section needs a bit of an overhaul. Have you taken a look at the react provider/context docs for inspiration at all? https://react.dev/learn/passing-data-deeply-with-context

docs/src/content/docs/basic-concepts.mdx
dereklegenzoff: I tested this link in dev and it worked so we should be all good.

examples/providers/README.md
dereklegenzoff: hmm interesting question. I think this is really a question on how users will use and look through examples and how we can optimize for that. 

Users could either:
a. Be looking for an example for a scenario to help implement something
b. Be looking to learn how to implement some pattern (structured outputs, providers)

My hypothesis is that b is more common and more useful for simple examples like this. Or maybe a different spin is that we want some examples that show a and others that show b

Web scraping alone is a good scenario to show although there's not much meat here on that front besides saying "you can use firecrawl"

docs/src/content/docs/concepts/providers.mdx
dereklegenzoff: sounds good, i will rev on this a bit

examples/providers/README.md
EvanBoyle: I'm fine with that.

docs/src/content/docs/concepts/providers.mdx
dereklegenzoff: added a bullet on this to the benefits section

docs/src/content/docs/concepts/providers.mdx
dereklegenzoff: Updated to this:

Providers offer several benefits:

- **Centralized Configuration Management** - providers manage settings like API keys and feature flags keeping configuration separate from business logic.
- **Data Sharing** - providers give you an easy way to share data between components without prop drilling to simplify your code.
- **Reusability** - providers and their associated components can easily be reused in different parts of your application.
- **Scoped State Management** - providers can be used to manage state for parts of your workflow or the entire thing.

dereklegenzoff:
Messed up this PR by accidentally rebasing so I'm just going to close this and open up a fresh PR.

PR #149, opened 2025-01-21
docs: adding openai component reference

dereklegenzoff:
Adds in the reference for openai components. Also trying to land on a good layout for reference pages


EvanBoyle (approved):
Looks great! My only suggestion is that we show at least one example of confiugring the provider with an openAI compatible endpoint like anthropic or perplexity. 

When Jeremy merges his PR for multiple provider support, we should probably have a section on using those patterns as well.

PR #152, opened 2025-01-21
style: updating all examples to use "start" instead of "run"

dereklegenzoff:
Updating all examples to use `start` instead of `run`. This aligns the examples with the output of `create gensx` and prevents the weirdness of `npm run run`


jmoseley (approved):
Nice!

Issue #150, opened 2025-01-21
Add gsx.array for map/reduce/filter operations over lists of elements


EvanBoyle:
Aimed at solving https://github.com/cortexclick/gensx/issues/107

Proposal is to add `gsx.array` and corresponding methods that match the javascript array stdlib like map, reduce, filter, etc. 

Every method like `map` would: 
1. execute the array to deeply resolve all components
2. call the transform function (ie a mapper) on the resolved array and return the values

Sketch of the implementation: 

```tsx
// Usage remains the same:
const workflow = gsx.array([1, 2, 3])
  .map(n => <NumberDoubler n={n} />)
  .filter(n => n > 4)
  .reduce((acc, n) => acc + n, 0);

// Implementation revised to use execute():
class GsxArray<T> {
  constructor(private items: T[]) {}

  async map<U>(fn: (item: T) => U): Promise<GsxArray<U>> {
    const resolvedItems = await execute<T[]>(this.items);
    const mappedItems = await Promise.all(
      resolvedItems.map(item => execute(fn(item)))
    );
    return new GsxArray(mappedItems);
  }

  async filter(predicate: (item: T) => boolean): Promise<GsxArray<T>> {
    const resolvedItems = await execute<T[]>(this.items);
    const filteredItems = await Promise.all(
      resolvedItems.filter(async item => await execute(predicate(item)))
    );
    return new GsxArray(filteredItems);
  }

  async reduce<U>(
    fn: (acc: U, item: T) => U,
    initial: U
  ): Promise<U> {
    const resolvedItems = await execute<T[]>(this.items);
    return resolvedItems.reduce(
      async (acc, item) => execute(fn(await acc, item)),
      Promise.resolve(initial)
    );
  }

  // Helper to get raw array
  async toArray(): Promise<T[]> {
    return execute<T[]>(this.items);
  }
}

const gsx = {
  // ... existing gsx methods ...
  
  array<T>(items: T[] | Promise<T[]>): GsxArray<T> {
    return new GsxArray(items);
  }
};
```

jmoseley:
I worry that this will shine a spotlight on the typescript issues that are not resolvable at this point.

Perhaps we should focus on just `gsx.map` for now to simplify the surface area, and rely on the native array operators for everything else?

EvanBoyle:
@jmoseley - you are worried that we won't get strong type inference on map/reduce/filter and will have to specify generic type params at every step?

jmoseley:
[...]

Issue #107, opened 2025-01-14
Define pattern for filtering and flattening lists


dereklegenzoff:
Currently, it can be tricky to handle scenarios where either a) you need to handle null items in a list or b) you need to flatten a list of lists. This is because the gensx elements haven't been resolved by the time you're running `.filter()` or `.flat()`.

Take the following example:

```typescript
<PromptToQuery prompt={prompt}>
        {({ queries }) => {
          return queries.map((query) => (
            <ArxivSearch query={query}>
              {(documents: ArxivEntry[]) => {
                console.log(`\n=== Results for query: "${query}" ===`);
                return documents.map((document) => (
                  <DocumentGrader prompt={prompt} document={document}>
                    {(gradeResult: DocumentGraderOutput) => {
                      console.log(
                        `Title: ${document.title}\nUseful: ${gradeResult.useful}\n`
                      );
                      return gradeResult.useful ? document : null;
                    }}
                  </DocumentGrader>
                )).filter((doc): doc is ArxivEntry => doc !== null); // <- this filter does not work
              }}
            </ArxivSearch>
          ));
        }}
</PromptToQuery>
```

The filter throws an error saying `A type predicate's type must be assignable to its parameter's type.Type 'ArxivEntry' is not assignable to type 'Element'.`

We should define a pattern for making it easy to filter and flatten list or, at a minimum, provide prescriptive guidance to users on how to handle this.

EvanBoyle:
One option is to add a set of gsx equivelents to `array.map` `array.filter`, etc. 

Roughly, these would just be syntactic sugar over `gsx.execute`:

```tsx
function map(items Element[], mapper: (item: any) => any){
 return await gsx.execute(items).map(mapper)
}

function filter(items Element[], filterFn: (item: any) => any){
 return await gsx.execute(items).filter(filterFn)
}
```

Need to get more concrete on the semantics, what it accepts/doesn't accept, and how the calls can be composed/chained together since they do produce promises.

PR #153, opened 2025-01-22
feat: Add `gsx.array` plus `map`, `filter`, `reduce`, and `flatMap`

EvanBoyle:
Adds a new `gsx.array` wrapper that exposes a fluent API for array operations over array elements that may contain components that need to be executed before execution. 

Fixes https://github.com/cortexclick/gensx/issues/150 fixes https://github.com/cortexclick/gensx/issues/107

Due to typescript type limitations, many of these methods do require a generic parameter that maps to the operations output type. Something we can fix up in the future via compiler extensions and IDE integration. 

An example: 

```tsx
  const arr = gsx.array<NumberWrapperOutput>([
    <NumberWrapper n={1} />,
    <NumberWrapper n={2} />,
    <NumberWrapper n={3} />,
    <NumberWrapper n={4} />,
    <NumberWrapper n={5} />,
  ]);
  const result = await arr
    .map<number>(n => <NumberDoubler value={n.value} />)
    .map<number>(n => <Value value={n} />)
    .filter((n: number) => <AsyncNumberFilter value={n} />)
    .reduce<number>((acc: number, n: number) => acc + n />, 0);
```

### DIFF
--- a/packages/gensx/tests/array.test.tsx
+++ b/packages/gensx/tests/array.test.tsx
@@ -1,0 +1,115 @@
+import { expect, test } from "vitest";
+
+import { gsx } from "../src";
+
+interface NumberWrapperOutput {
+  value: number;
+}
+
+const NumberWrapper = gsx.Component<{ n: number }, NumberWrapperOutput>(
+  "NumberWrapper",
+  ({ n }) => ({ value: n }),
+);
+
+const NumberDoubler = gsx.Component<{ value: number }, number>(
+  "NumberDoubler",
+  ({ value }) => value * 2,
+);
+
+const AsyncNumberFilter = gsx.Component<{ value: number }, boolean>(
+  "AsyncNumberFilter",
+  async ({ value }) => {
+    await new Promise(resolve => setTimeout(resolve, 1));
+    return value > 5;
+  },
+);
+
+const Value = gsx.Component<{ value: number }, number>(
+  "Value",
+  ({ value }) => value,
+);
+
+const TaskGenerator = gsx.Component<{ value: number }, number[]>(
+  "TaskGenerator",
+  async ({ value }) => {
+    // Simulate a workflow step that generates multiple tasks
+    await new Promise(resolve => setTimeout(resolve, 1));
+    // For each input, generate [value, value*2] as sub-tasks
+    return [value, value * 2];
+  },
+);
+
+test("map transforms array elements", async () => {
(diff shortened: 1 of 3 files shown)


packages/gensx/tests/array.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+const TaskGenerator = gsx.Component<{ value: number }, number[]>(
+  "TaskGenerator",
+  async ({ value }) => {
+    // Simulate a workflow step that generates multiple tasks
+    await new Promise(resolve => setTimeout(resolve, 1));
+    // For each input, generate [value, value*2] as sub-tasks
+    return [value, value * 2];
+  },
+);
+
```
jmoseley, line 41: I think we need tests for a component that uses `execute` inside of it, and a component that returns another component.

packages/gensx/tests/array.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+  ({ value }) => value,
+);
+
+const TaskGenerator = gsx.Component<{ value: number }, number[]>(
+  "TaskGenerator",
+  async ({ value }) => {
+    // Simulate a workflow step that generates multiple tasks
+    await new Promise(resolve => setTimeout(resolve, 1));
+    // For each input, generate [value, value*2] as sub-tasks
+    return [value, value * 2];
```
jmoseley, line 38: Test nesting?

```suggestion
    return [value, <NumberBoulder value={value} />];
```

jmoseley (approved):
Just need to add a bit more breadth to the tests, otherwise looks good.

packages/gensx/tests/array.test.tsx
EvanBoyle: Done.

packages/gensx/tests/array.test.tsx
EvanBoyle: 👍 updated the taskGenerator as suggested below to cover case (2) and will add a test for (1) as well.

EvanBoyle: Commit https://github.com/cortexclick/gensx/pull/153/commits/bef9c9d9ca5e4390c2721a2db81d5be4dc5caf20 adds tests for: 

1. Components that return components
2. Using gsx.array within a component
3. Using gsx.array within a child function.

PR #154, opened 2025-01-22
docs: adding docs and example on providers

dereklegenzoff:
Recreating #148 from a clean branch

Adds the concept doc for providers

Also adds a provider example that shows how to create a provider for Firecrawl (that is discussed in the doc)

It's worth noting that the top of this doc is the same as the text from the basic concepts page but it extends the info. Eventually we will also want an "Available Providers" section but when written currently, it only contains the OpenAIProvider.

### DIFF
--- a/examples/providers/README.md
+++ b/examples/providers/README.md
@@ -1,16 +1,8 @@
-# Reflection Example
+# Providers Example
 
-This example demonstrates how to use GenSX's `gsx.execute` to execute sub-workflows, as well as recursive component execution. It implements a buzzword cleaning system that uses GPT to iteratively remove business jargon while preserving the original meaning of the text.
+This example shows how to create a provider and a related component. In particular, it creates a provider for the [Firecrawl](https://www.firecrawl.dev/) API and uses it to scrape a page and return the markdown.
 
-## What it demonstrates
-
-- Recursive component implementation using `gsx.execute`
-- Integration with OpenAI's GPT model
-- State management through component props
-- Complex workflow with multiple steps:
-  1. Buzzword detection
-  2. Text cleaning via GPT
-  3. Recursive verification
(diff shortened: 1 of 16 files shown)


examples/providers/README.md
```diff
@@ -0,0 +1,28 @@
...
+pnpm install
+
+# Set your OpenAI API key
+export OPENAI_API_KEY=<your_api_key>
+
+# Run the example
+pnpm run start
+```
+
+The example will clean a sample text containing business buzzwords. You can modify the input text in `index.tsx` to clean different content.
```
EvanBoyle, line 28: Looks like this is a copy paste from the reflection example and needs an update?

examples/providers/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "scripts": {
+    "dev": "nodemon",
+    "start": "NODE_OPTIONS='--enable-source-maps' tsx ./index.tsx",
+    "build": "tsc",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "test": "NODE_OPTIONS='--enable-source-maps' tsx ./index.tsx"
```
EvanBoyle, line 15: Looks like we should remove this?

EvanBoyle (commented):
Two comments, but looks good to merge otherwise! (once we fix the build errors)

examples/providers/README.md
dereklegenzoff: ahh sorry about that, ended up manually creating this branch

examples/providers/package.json
dereklegenzoff: oh yeah, we should remove that from all the examples. pushed the changes

PR #155, opened 2025-01-22
feat: Add checkpoint support for StreamComponent

EvanBoyle:
Adds support for checkpointing within `StreamComponent`. Special care is taken to fork the stream so that we don't block. When the user reads the stream, we follow along, accumulate the value, and then send the final `completeNode` call with the complete value at the end. 

Fixes https://github.com/cortexclick/gensx/issues/142 fixes https://github.com/cortexclick/gensx/issues/136

### DIFF
--- a/packages/gensx/src/component.ts
+++ b/packages/gensx/src/component.ts
@@ -64,16 +64,63 @@ export function StreamComponent<P>(
   fn: (props: P) => MaybePromise<Streamable | JSX.Element>,
 ): GsxStreamComponent<P> {
   const GsxStreamComponent: GsxStreamComponent<P> = async props => {
-    const iterator: Streamable = await resolveDeep(fn(props));
-    if (props.stream) {
-      return iterator;
-    }
+    const context = getCurrentContext();
+    const workflowContext = context.getWorkflowContext();
+    const { checkpointManager } = workflowContext;
 
-    let result = "";
-    for await (const token of iterator) {
-      result += token;
+    // Create checkpoint node for this component execution
+    const nodeId = await checkpointManager.addNode(
+      {
+        componentName: name,
+        props: Object.fromEntries(
+          Object.entries(props).filter(([key]) => key !== "children"),
+        ),
+      },
+      context.getCurrentNodeId(),
+    );
+
+    try {
+      const iterator: Streamable = await context.withCurrentNode(nodeId, () =>
+        resolveDeep(fn(props)),
+      );
+
+      if (props.stream) {
+        // Create a wrapper iterator that captures the output while streaming
+        const wrappedIterator = async function* () {
+          let accumulated = "";
+          try {
+            for await (const token of iterator) {
+              accumulated += token;
+              yield token;
+            }
+            // Complete the checkpoint with accumulated output
+            checkpointManager.completeNode(nodeId, accumulated);
+          } catch (error) {
+            if (error instanceof Error) {
+              checkpointManager.addMetadata(nodeId, { error: error.message });
+              checkpointManager.completeNode(nodeId, accumulated);
+            }
+            throw error;
+          }
+        };
+        return wrappedIterator();
+      }
+
+      // Non-streaming case - accumulate all output then checkpoint
+      let result = "";
+      for await (const token of iterator) {
+        result += token;
+      }
+      checkpointManager.completeNode(nodeId, result);
+      return result;
+    } catch (error) {
+      // Record error in checkpoint
+      if (error instanceof Error) {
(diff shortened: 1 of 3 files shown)


packages/gensx/src/component.ts
```diff
@@ -64,16 +64,63 @@ export function StreamComponent<P>(
...
+        // Create a wrapper iterator that captures the output while streaming
+        const wrappedIterator = async function* () {
+          let accumulated = "";
+          try {
+            for await (const token of iterator) {
+              accumulated += token;
+              yield token;
+            }
+            // Complete the checkpoint with accumulated output
+            checkpointManager.completeNode(nodeId, accumulated);
```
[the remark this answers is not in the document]
jmoseley: This would have a side benefit of streaming to the console as well.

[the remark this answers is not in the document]
EvanBoyle: `completeNode` is mostly idempotent (though it does update the endTime) so it is safe to call multiple times.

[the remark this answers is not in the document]
EvanBoyle: There is benefit to preserving the actual endTime, so will push up another commit that does this.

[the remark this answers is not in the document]
EvanBoyle: @jmoseley fixed in https://github.com/cortexclick/gensx/pull/155/commits/775eab31014cbe1658e5f0f816a21e170e38392a

- endTime is preserved
- we record a `streaming in progress` output before the final output
- if there is an error in processing the stream we record that as well.

packages/gensx/src/component.ts
```diff
@@ -64,16 +64,74 @@ export function StreamComponent<P>(
...
-    let result = "";
-    for await (const token of iterator) {
-      result += token;
+        // Create a wrapper iterator that captures the output while streaming
+        const wrappedIterator = async function* () {
+          let accumulated = "";
+          try {
+            for await (const token of iterator) {
+              accumulated += token;
+              yield token;
```
jmoseley, line 97: We don't want to send updates for the intermediate values?

packages/gensx/tests/checkpoint.test.tsx
```diff
@@ -357,4 +362,107 @@ suite("checkpoint", () => {
       ],
     });
   });
+
+  test("handles streaming components", async () => {
```
jmoseley, line 366: I think this should be two tests.

packages/gensx/tests/checkpoint.test.tsx
```diff
@@ -357,4 +362,107 @@ suite("checkpoint", () => {
...
+
+    // Test streaming mode
+    const {
+      result: streamingResult,
+      checkpoints: streamingCheckpoints,
+      checkpointManager,
+    } = await executeWithCheckpoints<AsyncGenerator<string>>(
+      <StreamingComponent tokens={["Hello", " ", "World"]} stream={true} />,
+    );
+
```
jmoseley, line 405: We should verify the intermediate checkpoint state.

jmoseley, line 405: We should verify the intermediate checkpoint state.

jmoseley (approved):
Some comments on the tests, and one question. LGTM o/w.

packages/gensx/tests/checkpoint.test.tsx
EvanBoyle: I tried doing this, but there is no easy way to do so since intermediate checkpoints are intentionally skipped in the queue (and the skipping is somewhat non-deterministic depending on the timing of the event loop).

packages/gensx/tests/checkpoint.test.tsx
EvanBoyle: Good point, fixed via https://github.com/cortexclick/gensx/pull/155/commits/994e0a8016ad46d0ad0e138fc95c0a5dc0dc186d

EvanBoyle:
@jmoseley going to merge, but if you feel strongly about the intermediate streaming and state validation let me know and I can return to it in another PR.

packages/gensx/src/component.ts
jmoseley: Sounds good. I guess it would just add to that feeling of fast and reactive UX, to be able to watch it stream in while you are watching the workflow execute. I know I would have liked this in Langsmith for some of the long running streams we had.

Given the way we simply skip until the current update is done syncing, it will be extra writes, but not one for every token, or anything like that. But it might suggest moving to a debounced update model or something.

Happy to punt for now, but seems like a feature the reenforces that reactive experience we are going for. Maybe @danfhernandez has an opinion about how important it is.

PR #168, opened 2025-01-22
docs: adding concept doc for concepts

dereklegenzoff:
Adds a conceptual doc on concepts and a simple example. Also makes minor tweaks to the basic concepts section on contexts.


docs/src/content/docs/concepts/context.mdx
```diff
@@ -5,10 +5,61 @@ sidebar:
   order: 4
 ---
 
-GenSX is a simple framework for building complex generative AI and LLM workflows.
+Contexts provide a way to share data across components without explicitly passing them through props. They're similar to React's Context API but adapted for GenSX workflows.
 
-It uses `jsx` to define and orchestrate workflows with functional, reusable components:
+Contexts are particularly useful for:
+
+- Sharing configuration across multiple components
```
jmoseley, line 12: Might be worthwhile calling out examples such as client singletons.

jmoseley: I guess this is mentioned in the providers section below.

docs/src/content/docs/concepts/context.mdx
```diff
@@ -5,10 +5,61 @@ sidebar:
...
+- Providing dependencies to deeply nested components without prop drilling
+- Managing state that needs to be accessed by multiple components
+
+GenSX also supports [Providers](../providers) which are a specialized use of contexts that focus on managing configuration and dependencies for your workflow. If you're managing external service configurations like API keys or client instances, you should use providers instead of raw contexts.
```
jmoseley, line 16: This gives the impression that Providers aren't tightly coupled with context, when they really are. You can't (usefully) use a context without a provider.

examples/contexts/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  () => {
+    const user = gsx.useContext(UserContext);
+    return `Hello, ${user.name}!`;
+  },
+);
+
+async function main() {
+  // Provide a value to the context
+  const result = await gsx.execute(
+    <UserContext.Provider
```
jmoseley, line 26: Do you think its worth demonstrating nesting providers, or just leave that up to the docs?

examples/contexts/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "scripts": {
+    "dev": "nodemon",
+    "start": "NODE_OPTIONS='--enable-source-maps' tsx ./index.tsx",
+    "build": "tsc",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix"
+  },
+  "dependencies": {
+    "@gensx/openai": "workspace:*",
```
jmoseley, line 17:
```suggestion
```

dereklegenzoff: good catch. i also removed the regular openai package

docs/src/content/docs/concepts/context.mdx
dereklegenzoff: Very good point. I'll revise this. I probably also need to clean up some of the verbiage in the providers doc

examples/contexts/index.tsx
EvanBoyle: Agree we should probably show that you can nest providers and all values are available in the tree.

EvanBoyle (commented):
I think we need to clarify the difference between providers and context: 

- Context: defines what can be shared (a type and a default value)
- Provider: sets the actual value for the context and takes care of passing it through the component tree to all children
- `gsx.useContext`: how child components access context values from a provider

Is there a use case for using context without a provider?

Given all of this, I think it is worth asking ourselves whether it makes sense for there to be separate docs pages for these, or if we should just have a `Context & Providers` section.

jmoseley:
I think combined pages would be better. You can't have one without the other, they aren't standalone.

dereklegenzoff:
I was thinking something similar. I'll combine these into the same page.

examples/contexts/index.tsx
dereklegenzoff: added a short section in the docs

dereklegenzoff:
thanks for all of the feedback! Just pushed a combined version of the docs. Hopefully a decent improvement.

EvanBoyle (approved):
Overall, this looks good and I think we can ship it. 

There are two things we didn't cover here: 

1. Mutating context
2. When to use / not to use a context

For (1) I'm not sure how much there is to cover, and honestly we probably need to experiment a little bit more for ourselves to be certain. Thing could be tricky here as components are executed in parallel, so depending on the order of execution siblings may or may not see updates. I'm pretty sure that children will see updates. 

For (2) it might be worth noting that using contexts makes components in your workflow less reusable since they depend on global state. 

Honestly, I think we could leave both of these out for now until we have a little bit more experience ourselves, but thought I would mention.

docs/src/content/docs/concepts/contexts.mdx
```diff
@@ -0,0 +1,199 @@
+---
+title: Contexts and providers
```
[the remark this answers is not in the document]
dereklegenzoff: updated

docs/src/content/docs/concepts/contexts.mdx
```diff
@@ -0,0 +1,199 @@
...
+description: Learn how to use contexts and providers in GenSX to share data across components.
+sidebar:
+  order: 3
+---
+
+Contexts and providers are powerful tools in GenSX for sharing data and managing configuration across components without explicitly passing props through every level of your component tree. They work similarly to [React's Context API](https://react.dev/reference/react/useContext) but are adapted to work with GenSX workflows.
+
+## What are contexts and providers?
+
+Contexts and providers are two closely related concepts that work together to share data and manage dependencies across components.
```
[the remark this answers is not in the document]
dereklegenzoff: updated

PR #170, opened 2025-01-22
fix: Fix checkpoint race condition by handling out of order arrival of nodes

EvanBoyle:
The full graph is executed in parallel. This means that `addNode` calls don't necessarily occur in order, though usually they do. Luckily the context of the `parentId` is maintained correctly all the way through the execution tree.

The result is nodes that occassionally get dropped from the execution graph.

@danfhernandez was occasionally seeing instances where the first component was dropped from the checkpoint. I observed this once myself as well:

image

This change adds support for handling out of order insertion of nodes into the graph. If we try to insert a node with a parentId into the graph but the parent hasn't arrived yet, then we put it into an `orphaned` nodes list, and attempt to process it on subsequent `addNode` calls when other nodes arrive. In addition, we never write out a checkpoint when the tree is in an invalid state. 

I added tests that manually insert nodes into checkpoint manager to simulate multiple scenarios for out of order arrival to validate and lock in this behavior.

### DIFF
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -48,7 +48,7 @@ export interface CheckpointWriter {
 
 export class CheckpointManager implements CheckpointWriter {
   private nodes = new Map<string, ExecutionNode>();
-  private pendingRootId: string | null = null;
+  private orphanedNodes = new Map<string, Set<ExecutionNode>>();
   public root?: ExecutionNode;
   public checkpointsEnabled: boolean;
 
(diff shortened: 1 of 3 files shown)


packages/gensx/src/checkpoint.ts
```diff
@@ -1,16 +1,12 @@
 import { join } from "node:path";
 
 // Cross-platform UUID generation
-async function generateUUID(): Promise<string> {
+function generateUUID(): string {
   try {
     // Try Node.js crypto first
```
jmoseley, line 6:
```suggestion
```

packages/gensx/src/checkpoint.ts
```diff
@@ -69,15 +66,92 @@ export class CheckpointManager implements CheckpointWriter {
...
+      this.orphanedNodes.set(expectedParentId, orphans);
+    }
+    orphans.add(node);
+
+    // Add diagnostic timeout to detect stuck orphans
+    this.checkOrphanTimeout(node.id, expectedParentId);
+  }
+
+  private checkOrphanTimeout(nodeId: string, expectedParentId: string) {
+    setTimeout(() => {
```
jmoseley, line 89: Should you track this timeout and cancel/clean it up if we call if for the same `expectedParentId` multiple times?

EvanBoyle: It will log for each orphaned component. You would prefer to just log once per parent ID?

packages/gensx/tests/checkpoint.test.tsx
```diff
@@ -363,6 +374,111 @@ suite("checkpoint", () => {
     });
   });
 
+  test("handles out of order node insertion and tree reconstruction", () => {
```
jmoseley, line 377: can this be a test `suite`, with a few separate tests in it?

packages/gensx/src/checkpoint.ts
jmoseley: hmm, no I guess this is fine.

jmoseley (approved):
Just need to break apart the tests a bit, LGTM o/w.

packages/gensx/tests/checkpoint.test.tsx
EvanBoyle: Done via https://github.com/cortexclick/gensx/pull/170/commits/21a49cf14acc19a824ce38f3c587a63688f5b122

PR #173, opened 2025-01-23
feat: adding support for generic arrays in gsx.array

### LINKED ISSUE #171: gsx.array() should support vanilla arrays
Currently gsx.array() throws an error when you try to pass an array of vanilla objects.

For example, this code:

```tsx
        <PromptToQuery prompt={prompt}>
          {({ queries }) => {

            // Use the queries to fetch the documents
            const documents = gsx.array<string>(queries).map(
               (query) => {
                  return <ArxivSearch query={query}/>;
               }

          }}
        </PromptToQuery>
```

Throws the following error: *Argument of type 'string[]' is not assignable to parameter of type 'Element[]'.
  Type 'string' is not assignable to type 'Element'.ts(2345)*

dereklegenzoff:
Fixes #171 

Adds support so that you can pass in objects like a `string[]` to `gsx.array()`

### DIFF
--- a/packages/gensx/src/array.ts
+++ b/packages/gensx/src/array.ts
@@ -5,7 +5,7 @@ import { execute } from "./resolve";
 export class GsxArray<T> {
   private promise: Promise<unknown[]>;
 
-  constructor(items: JSX.Element[] | Promise<unknown[]>) {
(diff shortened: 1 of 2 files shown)


EvanBoyle (approved):
🚢

PR #180, opened 2025-01-24
docs: adding docs for `gsx.array`

dereklegenzoff:
Adding concept doc for `gsx.array`


docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+  order: 5
+---
+
+When building sophisticated LLM workflows, you'll often need to call the same component for each element in an array. Some example of this include:
+
+- Evaluating the usefulness of each chunk returned by your vector database
+- Translating a single document into multiple languages
+- Analyzing key themes in a set of customer reviews
+
+While you can accomplish this with standard GenSX components, it can lead to your components becoming more tightly coupled and less reusable. To solve this, GenSX supports `gsx.array` which allows you to chain together array operations like `map`, `filter`, `flatMap`, and `reduce`. The API follows the same conventions as arrays in JavaScript, making it familiar and intuitive to use.
```
EvanBoyle, line 14: I would make the intro language a little more plain here. 

```
GenSX includes an array primitive that lets you to map, reduce, filter, and operate over arrays of components with the same javascript standard library APIs. With it, you can chain array operations over lists of components and the framework takes care of executing components and producing outputs as needed. 

This is useful for tasks like grading lists of documents, or evaluating and filtering out LLM responses with low scores.
```

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+const wrappedNumbers = gsx.array([
+  <NumberWrapper n={1} />,
+  <NumberWrapper n={2} />,
+  <NumberWrapper n={3} />,
+]);
+```
+
+This shows a simplistic example of creating an array from components but in practice you can do this with any component, no matter how complex.
+
+To convert a `GsxArray` back to a plain array, you simply call the `toArray()` method.
```
EvanBoyle, line 81: should probably mention that `toArray` will produce an array of resolved component outputs.

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+// Create from components
+const wrappedNumbers = gsx.array([
+  <NumberWrapper n={1} />,
+  <NumberWrapper n={2} />,
+  <NumberWrapper n={3} />,
+]);
+```
+
+This shows a simplistic example of creating an array from components but in practice you can do this with any component, no matter how complex.
```
EvanBoyle, line 79: How about something like: 

This is a simple example of creating an array of components, but in practice you can chain any other complex operation like map, reduce, and filter.

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+To convert a `GsxArray` back to a plain array, you simply call the `toArray()` method.
+
+### Map and FlatMap
+
+`gsx.array` supports both `map` and `flatMap` operations. Both of these operations behave exactly as you'd expect if you're familiar with `array.map` and `array.flatMap` in JavaScript.
+
+#### Map
+
+Map is great for scenarios where you need to call a component for each element in the array:
```
EvanBoyle, line 89: I don't think we need to explain what map is. Maybe something simpler like "Map uses the same API as standard `array.map`, but the input array can contain components and the mapper function can return components as well:"

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+```tsx
+const numbers = gsx.array([1, 2, 3]);
+
+const doubled = await numbers.map((n) => <NumberDoubler value={n} />).toArray();
+// Result: [2, 4, 6]
+```
+
+#### FlatMap
+
+FlatMap is particularly useful when you have components that return arrays and you need to flatten the results back into a single array:
```
EvanBoyle, line 100: Similar here, I wouldn't focus on why it is useful. Let's just explain what it does.

EvanBoyle: Similar feedback for the intro sentence of the rest of the sections.

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+To break down this example a bit more:
+
+- `acc` is the accumulator, which starts as the initial value, in this case an empty string
+- `value` is the current section of the markdown content
+- `CombineSections` is a component that does the combining given both the accumulator and the current value
+
+## How array operations work
+
+When you boil it down, `gsx.array` is essentially a wrapper around the `gsx.execute` function.
```
EvanBoyle, line 200: I'd remove the "boil you it down". Maybe just: 

`gsx.array` is a conveinece wrapper around `gsx.execute` and the corresponding javascript array function. 

From there you could show the same `.map` code written via gsx.array and using gsx.execute to demonstrate.

EvanBoyle (commented):
Great start. I think we need to simplify a bunch of the language, and move this a little closer to a reference doc. IE assume that users know why working with arrays of stuff are useful and focus more instead on what it is and how to use it. That being said, having practical examples with links is great.

EvanBoyle:
Also, feels like we should make this accessible via the nav? It doesn't currently show up.

dereklegenzoff:
> Also, feels like we should make this accessible via the nav? It doesn't currently show up.

The only reason i haven't exposed this in the nav yet is because we have the skeleton pages that don't have any content yet. could just remove those for now.

> and move this a little closer to a reference doc
I wonder if maybe we should split this into two pages: a reference page and a concept page. To some extent we need both practical examples that show how to use this plus more of the nitty gritty details

EvanBoyle:
I would like to start publishing pages as they are ready, so let’s delete
the skeleton if it is getting in the way.

On your other note, I dont think we need either a full reference page or a
deep concepts doc on this. Something in between. It is pretty straight
forward and matches JS, not much new.

I think we could keep what we have here and just trim down the verbiage
related to particular use cases and call it done.

docs/src/content/docs/concepts/arrays.mdx
dereklegenzoff: i like it. adding.

docs/src/content/docs/concepts/arrays.mdx
dereklegenzoff: I like the first half of that (updated) but not the second half. This sentence is just about instantiating an array so the stuff on the back half feels out of place. I'm just trying to say "you can do this with any component" because the example is so contrived. 

How about something like `This is a simple example of creating an array of components, but in practice you can do this with any component.`

dereklegenzoff: Reading through it, i'm going to remove it. now that this reads more like a ref doc it doesn't really fit in.

docs/src/content/docs/concepts/arrays.mdx
dereklegenzoff: updated

docs/src/content/docs/concepts/arrays.mdx
dereklegenzoff: updated. Combined this verbiage with some of what was already there which seemed best

docs/src/content/docs/concepts/arrays.mdx
```diff
@@ -0,0 +1,228 @@
...
+  ({ value }) => {
+    return value % 2 === 0;
+  },
+);
+
+// Use a component that returns a boolean to filter
+const evenNumbers = await gsx
+  .array<number>([10, 11, 12, 13, 14])
+  .filter((n) => <EvenNumberFilter value={n} />)
+  .toArray();
```
EvanBoyle, line 138: This example feels a little contrived. I don't think we would actually suggest that anyone create an `EvenNumberFilter` component. Probably this example would feel a little more natural if it was something along the lines of: 

```tsx
const numbers = [<DoubleNumber n={1}/>, <DoubleNumber n={2}/>, <DoubleNumber n={3}/>];

gsx.array(numbers).filter(n => n < 5);
```

EvanBoyle (approved):
Overall looks good. I do worry a bit that most of the examples are toys, but it does make the code brief and easy to read. I think this is probably fine, but I do think the examples should feel practical. 

ie, the most common case is that you want to map/reduce/filter over an array of components that you've created in a previous step. One place in particular stuck out to me and I left a comment. But overall great starting point!

docs/src/content/docs/concepts/arrays.mdx
dereklegenzoff: This particular example is demonstrating that a component that returns a boolean can be used in a filter expression so we'd need an example that fit's that criteria. We could do something more complex but it seems important that we show the component code inline to get the point across for this particular one.

EvanBoyle: I'm fine leaving it as is, up to you. The example just struck me as a little odd, but we should probably just ship it and we can update it if we hear feedback from users that they have difficulty groking the concept.

dereklegenzoff: sounds good, will be easy to update in the future if needed.

PR #182, opened 2025-01-27
feat: Add secrets masking

EvanBoyle:
# Secrets Masking
This feature adds secrets masking to enable users to replace sensitive values in the checkpoint with a `[secret]` mask.

image

Fixes https://github.com/gensx-inc/gensx/issues/156

## Summary

- Add new `ComponentOpts` type
- Add support for selective or broad masking of props via property paths.
- Support broad masking of outputs via the "output" property path. 
- Support Components and StreamComponent
- Support transitive secrets masking across children
- Extensive tests (16 cases, 700 lines) across all the edge cases I could think of

## Implementation

Secrets can be masked by adding props property paths, or the "output" key to a new `ComponentOpts.secrets` array. 

Masked secrets are tracked by value, masking is transitive across children, and we search props and outputs to mask any value even if it just contains the masked value ( either `contains` or `===` will result in masking, but we are more selective with masking for contains). For instance if the secret is `secretsecretsecret` and a child component uses that value in a template string to output `this is a secretsecretsecret` the child output will be masked as `this is a [secret]`. 

We enforce this transitive relationship via context. We extended the parentId context that was added for checkpointing to include the entire node. This gives us access to the hierarchy of previously masked values. For every node that gets added to the graph, we walk it's ancestors looking for secret values that need to be masked in it's inputs and outputs. 

Value-based masking plus transitive child masking gives as much safety as possible, and minimizes the amount that a user needs to think about masking secrets when writing and composing component. 

## Using Secret Outputs

We've added a new `ComponentOpts` type. Components accept `defaultOpts` at declaration time, and 
users can specify a `componentOpts` prop at the time of instantiation to mark additional props and outputs as secrets.  

```tsx
// specify defaults at declaration time
export const OpenAIProvider = gsx.Component<ClientOptions, never>(
  "OpenAIProvider",
  (args) => {
    const client = new OpenAI(args);
    return <OpenAIContext.Provider value={{ client }} />;
  },
  {
    secrets: ["apiKey"],
  },
);

// specify additive secret props at the time you use a component. `apiKey` will still be secret in this case.
<OpenAIProvider apiKey={key} foo={...} componentOpts={secrets: ["foo"]}
```

To make a prop, a secret refer to it by property path in the array: 

```tsx
interface Inputs {
  foo: string;
  bar: { buzz: number };
  baz: number;
}

const secrets = ["bar.buzz", "baz"]
```

All outputs can be masked by adding the `output` path to the array. Selective output masking is not currently supported: 

```tsx
<Component componentOpts={secrets: ["output"]}
```

If the property path refers to a property that is an object or array, all nested values within that object or array will be masked as `[secret]`, but the keys and structure of the object/array will stay intact, primarily to help with debugging and visualization of outputs.

### DIFF
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -1,5 +1,7 @@
 import { join } from "node:path";
 
+import { ComponentOpts } from "./component";
+
 // Cross-platform UUID generation
 function generateUUID(): string {
   try {
@@ -32,6 +34,7 @@ export interface ExecutionNode {
     };
     [key: string]: unknown;
   };
+  componentOpts?: ComponentOpts;
 }
 
 export interface CheckpointWriter {
@@ -48,12 +51,26 @@ export interface CheckpointWriter {
 export class CheckpointManager implements CheckpointWriter {
   private nodes = new Map<string, ExecutionNode>();
   private orphanedNodes = new Map<string, Set<ExecutionNode>>();
+  private _secretValues = new Map<string, Set<unknown>>(); // Internal per-node secrets
+  private currentNodeChain: string[] = []; // Track current execution context
+  private readonly MIN_SECRET_LENGTH = 8;
   public root?: ExecutionNode;
   public checkpointsEnabled: boolean;
 
   // Track active checkpoint write
   private activeCheckpoint: Promise<void> | null = null;
   private pendingUpdate = false;
+
+  // Provide unified view of all secrets
+  get secretValues(): Set<unknown> {
+    const allSecrets = new Set<unknown>();
+    for (const secrets of this._secretValues.values()) {
+      for (const secret of secrets) {
+        allSecrets.add(secret);
+      }
+    }
+    return allSecrets;
+  }
 
   constructor() {
     // Set checkpointsEnabled based on environment variable
@@ -180,12 +197,16 @@ export class CheckpointManager implements CheckpointWriter {
       const baseUrl =
         process.env.GENSX_CHECKPOINT_URL ?? "http://localhost:3000";
       const url = join(baseUrl, "api/execution");
+
+      // Create a deep copy of the execution tree for masking
+      const maskedRoot = this.maskExecutionTree(structuredClone(this.root));
+
       const response = await fetch(url, {
         method: "POST",
         headers: {
           "Content-Type": "application/json",
         },
-        body: JSON.stringify(this.root),
+        body: JSON.stringify(maskedRoot),
       });
 
       if (!response.ok) {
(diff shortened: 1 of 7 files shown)


packages/gensx/src/checkpoint.ts
```diff
@@ -48,13 +51,27 @@ export interface CheckpointWriter {
 export class CheckpointManager implements CheckpointWriter {
   private nodes = new Map<string, ExecutionNode>();
   private orphanedNodes = new Map<string, Set<ExecutionNode>>();
+  private _secretValues = new Map<string, Set<unknown>>(); // Internal per-node secrets
+  private currentNodeChain: string[] = []; // Track current execution context
+  private readonly MIN_SECRET_LENGTH = 8;
```
jmoseley, line 56: Why does there need to be a minimum length?

jmoseley:
One general comment: I think we should move the configuration for output to its own key, to prevent ambiguity with a key containing a secret, called "output".

packages/gensx/src/checkpoint.ts
EvanBoyle: Minimums are common in masking implementations, otherwise you might mask an entire object and then end up with a masked value like "a" which has unintended side effects of masking a large fraction of your output. We should definitely keep this, but open to lowering the minimum value if we think 8 chars is too high.

EvanBoyle:
> One general comment: I think we should move the configuration for output to its own key, to prevent ambiguity with a key containing a secret, called "output".

I am open to this, but I was trying to keep the API a little cleaner. Otherwise we would end up with something like: 

```tsx
<Component componentOpts={secretProps: ["foo"], secretOutputs: ["bar"]}
```

We can certainly do this, but feels a bit messy IMO. The other thing we could do is use a special key like `gsxOutput`, `__output` or `@output`. What do you think @jmoseley?

jmoseley:
I'm not sure we need to support specific output keys, as opposed to simply a maskOutput configuration key that does the same thing as "output" does now.

Also, perhaps it should mask output by default, so this would only be used in a small number of cases?

EvanBoyle:
> Also, perhaps it should mask output by default, so this would only be used in a small number of cases?

I definitely don't think we should mask outputs by default. The common case is that this is an LLM output you want to debug in the trace. Masking these by default would make that very difficult. 

Just to be sure we're on the same page about what this means. Let's walk through what it means to mask something: 

1. You have a prop `foo: "supersecret"` 
2. You add `secrets: ["foo"]`
3. We capture the value of foo and add it to the secret values list `supersecret`
4. When we write the checkpoint, we look for any inputs and outputs that have that value, and mask it with `[secret]`
5. If the component happened to return an output `this is a secret: ${props.foo}` that resolved to `this is a secret: supersecret`, we would detect and mask this before writing the checkpoint resulting in `this is a secret: [secret]`. 

So secrets masking is really about collecting the set of sensitive values that should be removed from the props and output of this component and any of it's children. Does that make sense?

jmoseley:
Yes I believe we are on the same page. That is what I meant by secret masking, but I see the confusion.

So to clarify my proposal:
* secretProps config is used to identify values that are considered sensitive 
* Anywhere we see these sensitive values in either props or output, we mask them before sending them up as part of the checkpoint
* An extra config key, maybe "sensitiveOutput" or something like that indicates that the output from this component should be considered a sensitive value as well

This enables a component to produce sensitive output, but also have a prop named "output" and there not be ambiguity.

EvanBoyle:
I have a slight hesitation to split the API here as I like the simplicity of a single `secrets` array, but I guess it is worth being explicit and making sure there is no ambiguity. 

I will follow up with a slight variation of what you suggest: 
- `secretProps: string[]` for masking props
- `secretOutputs: boolean` for masking the entire output

Then if we ever have requests to support masking individual output keys, we can easily extend this API to be `secretOutputs: boolean | string[]` (and also extend the secretProps behavior to support the same).

jmoseley:
That sounds good!

packages/gensx/src/checkpoint.ts
```diff
@@ -199,6 +220,217 @@ export class CheckpointManager implements CheckpointWriter {
...
+    let result = value;
+
+    // Sort secrets by length (longest first) to handle overlapping secrets correctly
+    const secrets = Array.from(effectiveSecrets)
+      .filter(s => typeof s === "string" && s.length >= this.MIN_SECRET_LENGTH)
+      .sort((a, b) => String(b).length - String(a).length);
+
+    // Replace each secret with [secret]
+    for (const secret of secrets) {
+      if (typeof secret === "string") {
```
jmoseley, line 412: how do we get secrets that aren't strings?

packages/gensx/src/checkpoint.ts
```diff
@@ -199,6 +220,217 @@ export class CheckpointManager implements CheckpointWriter {
...
+    // Handle arrays and objects (excluding ArrayBuffer views)
+    if (Array.isArray(data) || !ArrayBuffer.isView(data)) {
+      const values = Array.isArray(data) ? data : Object.values(data);
+      values.forEach(value => {
+        this.collectSecretValues(value, nodeSecrets, visited);
+      });
+    }
+  }
+
+  private scrubSecrets(data: unknown, nodeId?: string, path = ""): unknown {
```
jmoseley, line 363: Is `path` used for anything? I don't see it here.

packages/gensx/tests/secrets.test.tsx
```diff
@@ -0,0 +1,696 @@
...
+      yield "not sensitive";
+    };
+    return stream();
+  },
+  { secrets: ["apiKey"] }, // Mark both input and output as secret
+);
+
+suite("tree reconstruction", () => {
+  beforeEach(() => {
+    process.env.GENSX_CHECKPOINTS = "true";
```
jmoseley, line 163: does this need to be reset in the `afterEach`?

packages/gensx/tests/secrets.test.tsx
```diff
@@ -0,0 +1,696 @@
...
+      yield apiKey;
+      yield "\nResponse: ";
+      yield "not sensitive";
+    };
+    return stream();
+  },
+  { secrets: ["apiKey"] }, // Mark both input and output as secret
+);
+
+suite("tree reconstruction", () => {
```
jmoseley, line 161: Is this suite related to secrets?

jmoseley (approved):

packages/gensx/src/checkpoint.ts
EvanBoyle: `scrubSecrets` is a recursive function that processes arrays and objects. When it hits a number or string, it tostrings it and calls this function. 

```ts
      if (typeof data === "string" || typeof data === "number") {
        const strValue = String(data);
        return this.scrubString(strValue);
      }
```

packages/gensx/src/checkpoint.ts
EvanBoyle: Yes, it is used for handling objects. 

```ts
      // Handle objects (excluding ArrayBuffer views)
      if (!ArrayBuffer.isView(data)) {
        return Object.fromEntries(
          Object.entries(data).map(([key, value]) => [
            key,
            this.scrubSecrets(value, nodeId, path ? `${path}.${key}` : key),
          ]),
        );
      }
```

packages/gensx/tests/secrets.test.tsx
EvanBoyle: We don't actually need to set this, so I removed it.

packages/gensx/tests/secrets.test.tsx
EvanBoyle: Accidental copy pasta fro the checkpoint tests. Removed.

PR #185, opened 2025-01-27
feat: initial work on tools

dereklegenzoff:
Currently this PR implements `gsx.Tool` and shows an [example](https://github.com/gensx-inc/gensx/pull/185/files#diff-74973ab4646a162de53e99155a939ff6d4d9c76819058d96d73f8451947bd8fe) of using it.

1/27 update:
- Pushed changes to wire up tools to the chat completion request
- Support responses in streaming and non-streaming context


packages/gensx/src/tools.ts
```diff
@@ -0,0 +1,63 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+      return config.function(validatedProps);
+    },
+  );
+
+  // Create a callable function that inherits from the component
+  const tool = Object.assign(component, {
+    schema: config.schema,
+    description: config.description,
+    function: config.function,
```
dereklegenzoff, line 46: this needs to call gsx.execute

PR #189, opened 2025-01-28
feat: Update checkpoint call to match the new API shape.

jmoseley:
## Proposed changes

Update the checkpoint persistence code to match the new API shape.

### DIFF
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -60,6 +60,9 @@ export class CheckpointManager implements CheckpointWriter {
   // Track active checkpoint write
   private activeCheckpoint: Promise<void> | null = null;
   private pendingUpdate = false;
+  private version = 1;
+  private org = "";
+  private apiKey = "";
 
   // Provide unified view of all secrets
   get secretValues(): Set<unknown> {
@@ -80,6 +83,25 @@ export class CheckpointManager implements CheckpointWriter {
       checkpointsEnv === "true" ||
       checkpointsEnv === "1" ||
       checkpointsEnv === "yes";
+
+    if (!this.checkpointsEnabled) {
+      return;
+    }
+
+    if (!process.env.GENSX_ORG) {
+      throw new Error(
+        "GENSX_ORG is not set, must be set to record checkpoints.",
+      );
+    }
+
+    this.org = process.env.GENSX_ORG;
+
+    if (!process.env.GENSX_API_KEY) {
+      throw new Error(
+        "GENSX_API_KEY is not set, must be set to record checkpoints.",
+      );
+    }
+    this.apiKey = process.env.GENSX_API_KEY;
   }
 
   private attachToParent(node: ExecutionNode, parent: ExecutionNode) {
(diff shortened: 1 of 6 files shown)


packages/gensx/src/checkpoint.ts
```diff
@@ -80,6 +83,25 @@ export class CheckpointManager implements CheckpointWriter {
...
+    }
+
+    this.org = process.env.GENSX_ORG;
+
+    if (!process.env.GENSX_API_KEY) {
+      throw new Error(
+        "GENSX_API_KEY is not set, must be set to record checkpoints.",
+      );
+    }
+    this.apiKey = process.env.GENSX_API_KEY;
```
EvanBoyle, line 104: I think we should adjust the way these flags compose now that we have an API key: 

1. presence of API key determines whether or not checkpoints are on
2. GENSX_CHECKPOINTS defaults to true, but setting it to false turns the feature off even if API is set. 

Feels weird to have to set: `GENSX_API_KEY=.. GENSX_CHECKPOINTS=true node ...`

packages/gensx/src/checkpoint.ts
```diff
@@ -194,22 +216,34 @@ export class CheckpointManager implements CheckpointWriter {
...
     try {
-      const baseUrl =
-        process.env.GENSX_CHECKPOINT_URL ?? "http://localhost:3000";
-      const url = join(baseUrl, "api/execution");
-
       // Create a deep copy of the execution tree for masking
       const maskedRoot = this.maskExecutionTree(structuredClone(this.root));
-
+      const baseUrl =
+        process.env.GENSX_CHECKPOINT_URL ?? "http://localhost:3001";
```
EvanBoyle, line 222: Very shortly we should change this default to be the production API. If this actually works in production, maybe we should change it now?

packages/gensx/src/checkpoint.ts
```diff
@@ -194,22 +216,34 @@ export class CheckpointManager implements CheckpointWriter {
...
+          completedAt: this.root.endTime,
+          rawExecution: JSON.stringify(maskedRoot),
+          // we leave this up to the client so that we save ourselves the cost of deserializing the tree on write.
+          steps,
+        }),
       });
 
-      if (!response.ok) {
+      if (response.ok) {
+        this.version++;
```
EvanBoyle, line 245: Why only increment version on success?

packages/gensx/tests/secrets.test.tsx
```diff
@@ -155,7 +155,16 @@ const SecretStreamComponent = gsx.StreamComponent<{ apiKey: string }>(
...
 suite("secrets", () => {
+  const originalEnv = {
+    GENSX_CHECKPOINTS: process.env.GENSX_CHECKPOINTS,
+    GENSX_ORG: process.env.GENSX_ORG,
+    GENSX_API_KEY: process.env.GENSX_API_KEY,
+  };
   beforeEach(() => {
+    process.env.GENSX_CHECKPOINTS = "true";
+    process.env.GENSX_ORG = "test-org";
+    process.env.GENSX_API_KEY = "test-api-key";
```
EvanBoyle, line 166: BTW - I seem to remember something about the timing of env vars meaning that we don't really have to set any of these values. Also because we're mocking out the fetch call in each of these. But if it is working, then nothing to worry about.

EvanBoyle (approved):
🚢

packages/gensx/src/checkpoint.ts
jmoseley: I like that, let me make that change.

packages/gensx/src/checkpoint.ts
jmoseley: Sure I can update it.

packages/gensx/src/checkpoint.ts
jmoseley: This will make it so the API receives monotonically increasing version numbers. But I guess that could lead to a bug if the server receives the update, but the client gets an error. I'll fix this.

packages/gensx/tests/secrets.test.tsx
jmoseley: Hmm. These tests were failing without them, but I can experiment.

jmoseley: I think we are going to end up in a world where we have weird test interactions, and maybe need to move to a model that does not directly rely on env vars (eg configuration for the execute call, that defaults to env vars, but we can set explicitly.

packages/gensx/tests/utils/executeWithCheckpoints.ts
```diff
@@ -0,0 +1,80 @@
...
+  element: ExecutableValue,
+): Promise<{
+  result: T;
+  checkpoints: ExecutionNode[];
+  checkpointManager: CheckpointManager;
+}> {
+  const checkpoints: ExecutionNode[] = [];
+
+  // Set up fetch mock to capture checkpoints
+  global.fetch = vi
```
EvanBoyle, line 33: Any worry that this sets up a moch and the caller has to know to tear it down? 

Also do we need to update tsconfig to make sure this doesn't end up in the bundle or does that happen automatically since this is in the test folder?

EvanBoyle (approved):
Looks good! 

Any work left here, or can we ship this and get main back into a functioning state against the deployed API?

packages/gensx/tests/utils/executeWithCheckpoints.ts
jmoseley: Tree shaking will handle the bundle problem.

Let me add resetAllMocks to a top level afterEach and that should cover it.

packages/gensx/tests/secrets.test.tsx
jmoseley: I think I fixed this.

PR #198, opened 2025-01-28
fix: Fix stream component type inference in child functions

EvanBoyle:
StreamComponent type inference was broken for nesting: 

```tsx
<ChatCompletion stream={true}>
 { (result) => {...}} // result is implicitly any...
</ChatCompletion>
```

This change improves the typing so that when `stream=false` result is automatically inferred as string, and when `stream=true` it is inferred as Streamable. 

Added tests that do nothing more than exercise the typescript compiler for this path. 

Fixes https://github.com/gensx-inc/gensx/issues/195

### DIFF
--- a/packages/gensx/src/types.ts
+++ b/packages/gensx/src/types.ts
@@ -67,28 +67,28 @@ export type Streamable =
   | AsyncIterableIterator<string>
   | IterableIterator<string>;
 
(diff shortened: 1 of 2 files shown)


jmoseley (approved):
Very nice.

PR #206, opened 2025-01-29
chore: Disable no-empty-object eslint rule.

jmoseley:
## Proposed changes

Disable the no-empty-object eslint rule so we can do things like `gsx.Component<{}, string>`


EvanBoyle (approved):
🚢

PR #211, opened 2025-01-31
feat: GSX ChatCompletion and OpenAIChatCompletion components

EvanBoyle:
WIP: Create a GSX ChatCompletion component that supports: 
- streaming
- tools
- structured outputs

All the types work nicely. Big key to making this play nicely was taking a dependency on `openai/helpers/zod` -- they expose functions for zod functions and zod -> JSON schema that is openAI compatible, and they maintain a fork of the zod-to-json-schema package that is tweaked to be compatible with OpenAI's internal type system. 

The `GSXCompletion` component also appropriately extends the OpenAI Completion input types. 

There is also a single `OpenAIChatCompletion` that supports both streaming / sync output against the full openAI SDK without any mucking about on our part. 

TODO: 
- This doesn't use OpenAI's `runTools` -- Does it need to? I'm not sure yet. 
- Maybe we can get rid of the `ChatCompletion` and `StreamCompletion` components, though might still be nice to have them. Might be nice to expose a `runTools` equivalent. Need to play around and make a decision. 
- Do we remove StreamComponent entirely? Looks like we can get around it with the type system alone. 
- GSX.Tool needs to be integrated with checkpointing - likely extend component somehow, and the tool function needs to be called via `gsx.execute` wrapper so that parenting and input/output capturing works properly.

### DIFF
--- a/packages/gensx-openai/src/composition.tsx
+++ b/packages/gensx-openai/src/composition.tsx
@@ -1,0 +1,244 @@
+import { gsx } from "gensx";
+import {
+  ChatCompletion as ChatCompletionOutput,
+  ChatCompletionChunk,
+  ChatCompletionCreateParamsNonStreaming,
+  ChatCompletionCreateParamsStreaming,
+  ChatCompletionMessageParam,
+} from "openai/resources/chat/completions";
+import { Stream } from "openai/streaming";
+import { z } from "zod";
+
+import { GSXStructuredOutput, GSXTool, OpenAIContext } from "./index.js";
+
+// Base types for raw OpenAI chat completion
+type RawCompletionProps = Omit<
+  ChatCompletionCreateParamsNonStreaming,
+  "stream" | "tools" | "response_format"
+>;
+
+type RawCompletionReturn = ChatCompletionOutput;
+
+// Raw completion component that directly calls OpenAI
+export const RawCompletion = gsx.Component<
+  RawCompletionProps,
+  RawCompletionReturn
+>("RawCompletion", async (props) => {
+  const context = gsx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  return context.client.chat.completions.create({
+    ...props,
+    stream: false,
+  } as ChatCompletionCreateParamsNonStreaming);
+});
+
+// Stream transform component
+type StreamTransformProps = RawCompletionProps & {
+  stream: true;
+  tools?: GSXTool<z.ZodObject<z.ZodRawShape>>[];
+};
+
+type StreamTransformReturn = Stream<ChatCompletionChunk>;
+
+export const StreamTransform = gsx.Component<
+  StreamTransformProps,
+  StreamTransformReturn
+>("StreamTransform", async (props) => {
+  const context = gsx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  const { stream, tools, ...rest } = props;
+  const openAITools = tools?.map((tool) => tool.toOpenAITool());
+
+  return context.client.chat.completions.create({
+    ...rest,
+    tools: openAITools,
+    stream: true,
+  } as ChatCompletionCreateParamsStreaming);
+});
+
+// Tool transform component
+type ToolTransformProps = RawCompletionProps & {
+  tools: GSXTool<z.ZodObject<z.ZodRawShape>>[];
+};
+
+type ToolTransformReturn = RawCompletionReturn;
+
+export const ToolTransform = gsx.Component<
+  ToolTransformProps,
+  ToolTransformReturn
+>("ToolTransform", async (props) => {
+  const context = gsx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  const { tools, ...rest } = props;
+  const openAITools = tools.map((tool) => tool.toOpenAITool());
+
+  return context.client.chat.completions.create({
+    ...rest,
+    tools: openAITools,
+    stream: false,
+  } as ChatCompletionCreateParamsNonStreaming);
+});
+
+// Structured output transform component
+type StructuredTransformProps<T> = RawCompletionProps & {
+  structuredOutput: GSXStructuredOutput<T>;
+  tools?: GSXTool<z.ZodObject<z.ZodRawShape>>[];
+};
+
+type StructuredTransformReturn<T> = T;
+
+export const StructuredTransform = gsx.Component<
+  StructuredTransformProps<unknown>,
+  StructuredTransformReturn<unknown>
+>("StructuredTransform", async (props) => {
+  const context = gsx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  const { structuredOutput, tools, ...rest } = props;
+  const openAITools = tools?.map((tool) => tool.toOpenAITool());
+
+  const completion = await context.client.chat.completions.create({
+    ...rest,
+    tools: openAITools,
+    response_format: structuredOutput.toResponseFormat(),
+    stream: false,
+  } as ChatCompletionCreateParamsNonStreaming);
+
+  const content = completion.choices[0].message.content;
+  if (!content) {
+    throw new Error("No content returned from OpenAI");
+  }
+
+  try {
+    const parsed = JSON.parse(content) as unknown;
+    const validated = structuredOutput.safeParse(parsed);
+    if (!validated.success) {
+      throw new Error(`Invalid structured output: ${validated.error.message}`);
+    }
+    return validated.data;
+  } catch (e) {
+    throw new Error(
+      `Failed to parse structured output: ${e instanceof Error ? e.message : String(e)}`,
+    );
+  }
+});
+
+// Retry transform component for structured output
+type RetryTransformProps<T> = StructuredTransformProps<T> & {
+  maxAttempts?: number;
+};
+
+export const RetryTransform = gsx.Component<
+  RetryTransformProps<unknown>,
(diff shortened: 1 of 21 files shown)


packages/gensx-openai/src/index.tsx
```diff
@@ -1,18 +1,207 @@
...
+
+    try {
+      const parsed = JSON.parse(content) as unknown;
+      const validated = structuredOutput.safeParse(parsed);
+      if (!validated.success) {
+        throw new Error(
+          `Invalid structured output: ${validated.error.message}`,
+        );
+      }
+      return validated.data;
```
danfhernandez, line 186: Would we want to use the structured output and chat completions components here?

packages/gensx-openai/src/index.tsx
```diff
@@ -1,18 +1,207 @@
...
+  // Handle streaming case
+  if (props.stream) {
+    const { tools, ...rest } = props;
+    const openAITools = tools?.map((tool) => tool.toOpenAITool());
+
+    const response = await context.client.chat.completions.create({
+      ...rest,
+      stream: true,
+      tools: openAITools,
+    } as ChatCompletionCreateParamsStreaming);
```
danfhernandez, line 157: Why not use the chat completions component and pipe it to streamable?

packages/gensx-openai/src/index.tsx
```diff
@@ -1,18 +1,207 @@
...
+
+  // Handle standard case (with or without tools)
+  const { tools, ...rest } = props;
+  const openAITools = tools?.map((tool) => tool.toOpenAITool());
+
+  return context.client.chat.completions.create({
+    ...rest,
+    stream: false,
+    tools: openAITools,
+  } as ChatCompletionCreateParamsNonStreaming) as Promise<ChatCompletionOutput>;
```
danfhernandez, line 202: Again why not just return the chat completion component here?

danfhernandez:
> This doesn't use OpenAI's runTools -- Does it need to? I'm not sure yet.
Does this work if it's not openai? I was thinking we'd reimplement but if it works... perhaps it's worth considering

> Maybe we can get rid of the ChatCompletion and StreamCompletion components, though might still be nice to have them.
Might be nice to expose a runTools equivalent. Need to play around and make a decision.
I think we should keep ChatCompletion at least 

>GSX.Tool needs to be integrated with checkpointing - likely extend component somehow, and the tool function needs to be called via gsx.execute wrapper so that parenting and input/output capturing works properly.
Is there a reason this can't just be a normal component? I'm probably missing something. I would think the tool function is a component itself.

EvanBoyle:
@danfhernandez I think we should probably support the monolithic "chatCompletion" component as implemented here, but also support the individual <ChatCompletion>, <Streamable>, <StructuredOutputs>, <RunTools>, etc, as well. 

This rough cut was just proving to myself that the type system could support the monolithic use case, and that we could get rid of streamcomponent altogether.

danfhernandez:
> GSX.Tool needs to be integrated with checkpointing - likely extend component somehow, and the tool function needs to be called via gsx.execute wrapper so that parenting and input/output capturing works properly.

Gotcha, looks good there! 

> @danfhernandez I think we should probably support the monolithic "chatCompletion" component as implemented here, but also support the individual <ChatCompletion>, <Streamable>, <StructuredOutputs>, <RunTools>, etc, as well.

I mean I guess I just wonder why not reuse these under the hood even just from an observability perspective? Especially if it's going to be very similar code

EvanBoyle:
@danfhernandez both tools and structured output in this example are pretty tightly coupled to the OpenAI SDK call. 

Do you think there is opportunity to decouple and decompose into components here? I am uncertain.

EvanBoyle:
I guess as currently implemented. This does not execute the tools. It will just return back the tool calls that need to be made.

EvanBoyle:
@danfhernandez updated the PR to have an alternative approach in `composition.tsx`. This is pretty cool because it does: 

- tool calling via composition (generate tool calls -> execute tool calls -> execute final completion with results)
- structured outputs with retries via composition (generate the structured output tool call, execute it, if  the parsing fails, try again automatically up to a limit)

packages/gensx-openai/src/composition.tsx
```diff
@@ -0,0 +1,359 @@
...
+  tools: GSXTool<z.ZodObject<z.ZodRawShape>>[];
+  toolCalls: NonNullable<
+    ChatCompletionOutput["choices"][0]["message"]["tool_calls"]
+  >;
+  messages: ChatCompletionMessageParam[];
+}
+
+type ToolExecutorReturn = ChatCompletionOutput;
+
+export const ToolExecutor = gsx.Component<
```
danfhernandez, line 80: I wonder if naming should be nouns or verbs? 

I kind of like verbs for things like this. RunTool perhaps?

packages/gensx-openai/src/composition.tsx
danfhernandez, line 1: Is this another case where your goal is to demonstrate the inputs and outputs and not implementation?

packages/gensx-openai/src/composition.tsx
```diff
@@ -0,0 +1,359 @@
...
+});
+
+// Tool transform component
+type ToolTransformProps = RawCompletionProps & {
+  tools: GSXTool<z.ZodObject<z.ZodRawShape>>[];
+};
+
+type ToolTransformReturn = RawCompletionReturn;
+
+export const ToolTransform = gsx.Component<
```
danfhernandez, line 138: I think this transform naming is throwing me off. Doesn't this actually run the tools if you use it?

packages/gensx-openai/src/composition.tsx
```diff
@@ -0,0 +1,359 @@
...
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  const { stream, tools, ...rest } = props;
+  const openAITools = tools?.map((tool) => tool.toOpenAITool());
+
+  return context.client.chat.completions.create({
```
danfhernandez, line 62: Should this return rawcompletion?

packages/gensx-openai/src/composition.tsx
```diff
@@ -0,0 +1,359 @@
...
+          }`,
+        );
+      }
+    }),
+  );
+
+  // Add tool results to messages and make another completion
+  const updatedMessages = [...messages, ...toolResults];
+
+  const completion = await context.client.chat.completions.create({
```
danfhernandez, line 123: Could this return <RawCompletion />?

packages/gensx-openai/src/composition.tsx
```diff
@@ -0,0 +1,359 @@
...
+>("ToolExecutor", async (props) => {
+  const { tools, toolCalls, messages } = props;
+  const context = gsx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
+
+  // Execute each tool call
```
danfhernandez, line 92: I wonder if there's a way to componentize this

packages/gensx-openai/src/index.tsx
```diff
@@ -1,68 +1,245 @@
...
+    });
 
-      return generateTokens();
-    }
+type OpenAIChatCompletionReturn<P> = P extends { stream: true }
+  ? Stream<ChatCompletionChunk>
+  : ChatCompletionOutput;
+
+// Direct wrapper over OpenAI's chat completion
+export const OpenAIChatCompletion = gsx.Component<
```
dereklegenzoff, line 220: I like the new naming of `OpenAIChatCompletion` and `ChatCompletion`. I was just thinking that each package exporting it's own thing named `ChatCompletion` was going to get very confusing.

danfhernandez: > I like the new naming of OpenAIChatCompletion and ChatCompletion. I was just thinking that each package exporting it's own thing named ChatCompletion was going to get very confusing.
Agree! I like this naming as well

danfhernandez: Hmm in theory they shouldn't have the same output type right?

dereklegenzoff:
Just spent some time looking at the component naming again. I think it's probably worth bikeshedding on the naming of the components before we merge this to make sure we get it right. Some feedback from me:

- I'd vote we call it `GsxChatCompletion` rather than `GsxCompletion`. 
   - Just saying completion makes me think its a standard completions as opposed to chat completions which was how LLMs (gpt-2, gpt-3, etc) used to work back before ChatGPT came out and everyone adopted the chat format. 
   - Some open source models still support standard completion interfaces (llama2 for example) although this is increasingly antiquated. I wouldn't be all that surprised if eventually there's a need to support standard completions though.
- I have a similar aversion to the name `TextCompletion`; makes me think of standard completions. Could `GsxChatCompletion` just output a string if no tools are specified?
- I like the syntax of `gsx.tool() better than `new GsxTool()` because it aligns better with `gsx.array()`. `GsxTool` also reads exactly like `GsxStucturedOutput` which feels mismatched since the latter is actually making the request to the LLM.

packages/gensx-openai/src/newCompletion.tsx
```diff
@@ -0,0 +1,101 @@
...
+              ? {
+                  type: "string",
+                  description: (value as z.ZodString).description,
+                }
+              : { type: "string" },
+          ]),
+        ),
+        required: Object.keys(this.parameters.shape),
+      },
+    };
```
dereklegenzoff, line 74: Ran into this where everything is being converted to a string so you can't have tools that support other types. 

When i implemented this, I used `import { zodFunction } from "openai/helpers/zod";` to go from the zod schema to `ChatCompletionTool` format. 

```ts
    this.function = zodFunction({
      name: this.name,
      parameters: this.parameters,
      description: this.description,
    });
 ```
 
 Tried making the update here and caused some errors so it's not quite a trivial update

Comment (EvanBoyle) [reply packages/gensx-openai/src/newCompletion.tsx:74]: Will take a look. Thanks for pointing out.

Comment (dereklegenzoff) [reply packages/gensx-openai/src/newCompletion.tsx:74]: actually the super easy fix is this

```
    this.function = {
      name: this.name,
      description: this.description,
      parameters: zodToJsonSchema(this.parameters),
    };
```

`zodToJsonSchema` is the same thing that OpenAI's `zodFunction` helper uses under the hood.

EvanBoyle:
Closed in favor of https://github.com/gensx-inc/gensx/pull/237

PR #212, opened 2025-01-31
feat: CLI Device auth flow

jmoseley:
## Proposed changes

* Bones of the CLI package
* Implement device auth flow to obtain an API key from the API via browser authentication


EvanBoyle:
😍 

First question: how do users install it? What command do you run to invoke it? (I would really prefer to have an option available to install this and make it globally available in the shell as `gensx`, rather than making users type out `gensx-cli`).

jmoseley:
If you `npm i --global @gensx/cli` you will get `gensx-cli` on your path. We can update that to `gensx` easily.

We will also be able to support `npx gensx` if we reexport this package via the `gensx` package.

jmoseley:
Ok now this _should_:

* Support `npm install --global @gensx/cli` and then you can call `gensx login`
* Also work for `npx gensx login`

jmoseley:
I'm pretty sure its not a stretch to get this into `brew` as well. Just need to find an example we can copy of someone else that does that same thing.

jmoseley:
Added https://github.com/gensx-inc/homebrew-gensx/tree/v0 as a PoC for installing with brew. It works, just need to add a CI task that updates the release automatically.

packages/gensx-cli/src/commands/login.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    const verificationCode = generateVerificationCode();
+    const request = await createLoginRequest(verificationCode);
+    spinner.succeed();
+
+    const authUrl = new URL(`/auth/device/${request.requestId}`, APP_BASE_URL);
+    authUrl.searchParams.set("code_verifier", verificationCode);
+
+    consola.box(
+      `Press any key to open your browser and authenticate with GenSX:
```
EvanBoyle, line 179: Let's go ahead and ship this, but we should probably add a fallback incase this fails that let's users paste in an access code. Maybe open a follow up issue?

EvanBoyle (approved):

packages/gensx-cli/src/commands/login.ts
jmoseley: Good point. It does print the URL, but it is not polling at that point.

jmoseley: Unfortunately we need the API key _and_ org....

EvanBoyle: Once we have an API key, we can poll for available orgs. 

If there is already an org in the config file, we could just skip writing one.

PR #217, opened 2025-01-31
docs: adding concept doc on how gensx works

dereklegenzoff:
Adds a concept doc describing how gensx works


dereklegenzoff:
replacing with #277

Issue #74, opened 2025-01-07
Support multiple providers


jmoseley:
Currently, we use a provider to append data to the context, so that subsequent components can access it.

This is used by `@gensx/openai` to pass around the OpenAI client.

```
<OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
  <ChatCompletion model="gpt-4o-mini" messages={[ role: 'user', content: 'hello' ]} />
</OpenAIProvider>
```

However, this breaks down if you would like to use multiple instances of the OpenAI client (eg one for OpenAI, another for perplexity). We need to support a pattern for controlling how `ChatCompletion` (or some other context dependent component) uses the context and accesses its dependencies. For LLMs based components, there is a bit of additional complexity, since we likely want to specify both an SDK instance, and a model.

This gets more complicated when we imagine more complex components or workflows being shared, where you may need to be able to inject multiple different versions of the OpenAI SDK, and specify models along with it.

The scenarios we should consider:

1. I'm writing something locally and I have control over all of the code
2. I'm using a component written by someone else
3.  everything within a component uses one LLM provider
4. component might use different LLM providers (claude for some, openai for others)
5. my program uses both openai and perplexity

EvanBoyle:
So it turns out we already support all of the use cases we need. Here is the pattern for using multiple providers within a single program where all components are locally defined: 

```tsx

// make no changes, for local components

<OpenAIProvider>
    <SummarizationComponent />
</OpenAIProvider>

<PerplexityProvider>
    <MyWebResearchComponent />
</PerplexityProvider>

```

And here is the pattern for writing and using a third-party component that is published to NPM:

```tsx
// implementation:
interface ProviderConfig {
  clientOptions: openai.ClientOptions
  defaultModel: string
}

interface SharedComponentProps {
  inputString
  defaultProvider: ProviderConfig;
  webResearchProvider?: ProviderConfig;
  smallModelProvider?: ProviderConfig;
}

const SharedComponent = gsx.Component<SharedComponentProps, string>("SharedComponent", (props) => {
  return (
   <OpenAIProvider ...props.defaultProvider.clientOptions/>
      <ChatCompletion model={props.defaultProvider.model} .../>
   </OpenAIProvider>
  );

})


// usage:

const clientOptions: openai.ClientOptions = { apiKey: process.env.OPENAI_API_KEY, serverUrl: "..."};
const model = "gpt-4o-mini";

return (
[...]

PR #228, opened 2025-02-07
docs: adding new provider examples

dereklegenzoff:
- Adds docs and an example on using multiple OpenAI providers in the same workflow
- Adds docs and an example on creating reusable, sharable components

I originally tried to just add the new examples to the existing providers example but it felt like it got overloaded so ended up splitting it into three separate bite-sized examples.

Related to #74


docs/src/content/docs/concepts/context.mdx
```diff
@@ -188,9 +188,53 @@ In this example, the `WebPageSummarizer` component can access both the OpenAI cl
 
 The order of nesting doesn't matter as long as the component using a context is wrapped by its corresponding provider somewhere up the tree.
 
-## Additional Resources
+### Using multiple OpenAI compatible providers
+
+You can also nest multiple providers to use LLMs from different labs in the same workflow. When you nest multiple OpenAIproviders, components will always use the closest provider in the tree. The example below shows how you could use models from OpenAI and Groq in the same workflow using their OpenAI-compatible APIs.
```
jmoseley, line 193: Nit: not sure if it's worth clarifying closest provider vs closest provider above it in the tree.

dereklegenzoff: updated

EvanBoyle (approved):

PR #231, opened 2025-02-12
fix: Separate `children` from other props before component execution

EvanBoyle:
Fixes https://github.com/gensx-inc/gensx/issues/226

The root cause of that bug is passing props from a parent component to a child that runs via `gsx.execute`: 

```tsx
const Parent = gsx.Component("Parent", (props) => {
  // boom - this blows up because child is unexpectedly receiving `props.children`
  gsx.execute(<Child ...props/>);

  return { somethingElse }
}
```

This change fixes that with a few changes: 
- separate out children from the rest of props before a component function gets called/executed
- rework the parentID execution context flow to make sure that checkpointing and provider context continues to flow as expected

Testing: 
- Added a test to make sure that a component's function does not receive `props.children`
- Fixed an errant component renaming test that was trying to render children (we should never do this)

### DIFF
--- a/packages/gensx/src/context.ts
+++ b/packages/gensx/src/context.ts
@@ -101,7 +101,13 @@ export class ExecutionContext {
   }
 
   withCurrentNode<T>(nodeId: string, fn: () => Promise<T>): Promise<T> {
-    return withContext(this.withContext({ [CURRENT_NODE_SYMBOL]: nodeId }), fn);
+    return withContext(
+      this.withContext({ [CURRENT_NODE_SYMBOL]: nodeId }),
+      async () => {
+        const result = await fn();
+        return result;
+      },
+    );
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/gensx/src/context.ts
```diff
@@ -101,7 +101,13 @@ export class ExecutionContext {
   }
 
   withCurrentNode<T>(nodeId: string, fn: () => Promise<T>): Promise<T> {
-    return withContext(this.withContext({ [CURRENT_NODE_SYMBOL]: nodeId }), fn);
+    return withContext(
+      this.withContext({ [CURRENT_NODE_SYMBOL]: nodeId }),
+      async () => {
```
jmoseley, line 106: Is this extra anonymous function necessary?

jmoseley (approved):

packages/gensx/src/context.ts
EvanBoyle: Good catch, fixed.

PR #242, opened 2025-02-14
feat: Removing GSXSchema, updating docs, and updating example

dereklegenzoff:
This change includes the following:
- Removes `GSXSchema` in favor of using Zod Schemas directly
- Adds a couple of basic tests for structured outputs
- Updates example
- Updates docs

### DIFF
--- a/docs/src/content/docs/patterns/structured-outputs.mdx
+++ b/docs/src/content/docs/patterns/structured-outputs.mdx
@@ -5,35 +5,43 @@ sidebar:
   order: 4
 ---
 
-Workflows regularly require getting structured outputs (JSON) from LLMs. This guide shows how to use structured outputs with GenSX and OpenAI models but you can follow a similar pattern with other LLM providers.
+Workflows regularly require getting structured outputs (JSON) from LLMs. GenSX supports an easy way to get structured outputs from an LLM by using the `GSXChatCompletion` component and Zod. However, you can also use the `response_format` prop to get a structured output from OpenAI models without using GenSX helpers, just like you would when using the OpenAI SDK or API.
 
-All GenSX components include an input and output type. The most common use case is to use structured outputs to ensure that the response from the LLM matches the schema of the component's output type.
+This guide shows how to use structured outputs both with and without the helpers that GenSX provides. You can find the complete example code in the [structured outputs example](https://github.com/gensx-inc/gensx/tree/main/examples/basicExamples/structuredOutputs/index.tsx).
 
-## Basic usage
+## Structured outputs with GSXChatCompletion
 
-The most basic way to get a structured output from an LLM is to use the `response_format={{ type: "json_object" }}` option and then parse the response in the child function as shown below.
+When using the `GSXChatCompletion` component, you can define the Zod schema for the output format you want and GenSX will handle the rest.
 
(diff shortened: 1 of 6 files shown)


docs/src/content/docs/patterns/structured-outputs.mdx
```diff
@@ -47,51 +55,53 @@ const ExtractEntitiesBasic = gsx.Component<
...
+<GSXChatCompletion
+  model="gpt-4o-mini"
+  messages={[
+    {
+      role: "user",
+      content: prompt,
+    },
+  ]}
+  outputSchema={ExtractEntitiesSchema}
+  retry={{
```
EvanBoyle, line 86: What is the default value here? I didn't think through this much when originally implementing, we should set something intentional.

packages/gensx-openai/src/structured-output.tsx
```diff
@@ -88,7 +55,7 @@ export const StructuredOutput = gsx.Component<
           {...rest}
           messages={messages}
           tools={tools}
-          response_format={outputSchema.toResponseFormat()}
+          response_format={zodResponseFormat(outputSchema, "output_schema")}
```
EvanBoyle, line 58: Do we have an issue open to make this work without response_format for non-openai compatible models? 

This is the last point of the design that I'm concerned about. (I'm also concerned about what logic we use to detect this, since we should still use response_format for azure hosted models as well).

packages/gensx-openai/tests/structured-output.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  test("structured output works with `GSXChatCompletion` component", async () => {
+    const schema = z.object({
+      name: z.string(),
+      age: z.number(),
+    });
+
+    const TestComponent = gsx.Component("TestComponent", () => (
+      <GSXChatCompletion
+        model="gpt-4o"
```
EvanBoyle, line 89: Let's use cheap models (4o-mini) for tests.

EvanBoyle (approved):
Looks good. A few comments mostly unrelated to the core change.

packages/gensx-openai/tests/structured-output.test.tsx
dereklegenzoff: this is mocked so it's not actually calling the model

docs/src/content/docs/patterns/structured-outputs.mdx
dereklegenzoff: maxAttempts defaults to 3 meaning it will retry twice which seems reasonable to me.

packages/gensx-openai/src/structured-output.tsx
dereklegenzoff: I think the higher priority would be to make other models support the outputSchema parameter. I'll open issues. I'm not sold on the fact that we should make `response_format` work for non-openai models. We can't support json mode so it would end up just being a subset of the response_format functionality. Plus users should really use the helper we define.

Regarding the logic, I think we should look at the baseURL set in the client (just need to check for the openai and azure openai base urls hopefully).

dereklegenzoff: #246

PR #251, opened 2025-02-18
feat: first round of tool updates

dereklegenzoff:
Makes the following updates:
- Adds some basic tests for tools
- Fixes limitation around schemas only supporting strings. using `zodToJSONSchema` following the same pattern from an openai [example](https://github.com/openai/openai-node/blob/master/examples/function-call-helpers-zod.ts). 
- Fixes display in checkpoints to not display all the extra fluff from zod and gsxtool. Note that `GSXChatCompletion` still shows the extra fluff, will need to sort that out separately
- Updated the output type of `ToolExecutor` to be ChatCompletionToolMessageParam which is more specific and makes dealing with the outputs easier. 

Still have other updates to tools i need to make but wanted to batch the first round of changes together.

## Example tool definition

```ts
  const testTool = new GSXTool({
    name: "test_tool",
    description: "A test tool",
    schema: testToolSchema,
    execute: async (args) => {
      await Promise.resolve();
      return `Processed: ${args.input}`;
    },
  });
```

## Tool view in ux
Output now looks like this:
image

whereas before it looked like this
image


examples/gsxChatCompletion/index.tsx
```diff
@@ -46,18 +41,18 @@ async function tools() {
...
-  const tool = new GSXTool<typeof weatherSchema>(
-    "get_weather",
-    "get the weather for a given location",
-    weatherSchema,
-    async ({ location }: WeatherParams) => {
+  const tool = new GSXTool<typeof weatherSchema>({
+    name: "get_weather",
+    description: "get the weather for a given location",
+    schema: weatherSchema,
+    execute: async ({ location }: WeatherParams) => {
```
EvanBoyle, line 48: Two things now that I'm looking at this code for the third time: 

(1) We have both `<typeof weatherSchmea>` and `schema: weatherSchema`>. I haven't tried this, but maybe it is possible to remove the generic declaration? IE users just specify `new GSXTool({...`? Probably worth a try to see if we can get it to work. 

(2) I'm not in love with the word `execute` here. I wonder if we could come up with something else? OpenAI uses [`function` in runToools](https://github.com/openai/openai-node/blob/master/helpers.md#automated-function-calls) but I am not in love with using a reserved keyword for a parameter name. How about `fn`?

packages/gensx-openai/src/tools.tsx
```diff
@@ -9,51 +9,48 @@ import {
...
+    this.description = params.description;
+    this.schema = params.schema;
+    this.options = params.options ?? {};
+
+    this.definition = {
+      type: this.type,
+      function: {
+        name: this.name,
+        description: this.description,
+        parameters: zodToJsonSchema(this.schema),
```
EvanBoyle, line 45: You feel good about using the standard `zodToJsonSchema` rather than the vendored fork that OpenAI maintains? I am slightly worried as they probably have a fork for some compat reasons with their API.

packages/gensx-openai/src/stream.tsx
```diff
@@ -69,6 +73,10 @@ export const StreamCompletion = gsx.Component<
 
   // No tools, just stream normally
   return gsx.execute<Stream<ChatCompletionChunk>>(
-    <OpenAIChatCompletion {...rest} tools={tools} stream={true} />,
+    <OpenAIChatCompletion
+      {...rest}
+      tools={tools?.map((t) => t.definition)}
```
EvanBoyle, line 78: Is this what cleans up the output and removes all of the extraneous functions from the trace view?

EvanBoyle (approved):
Overall looks good! I think we can merge this as is, but a few small comments. It would be great if we could remove the need to specify a generic on GSX.Tool. Feels like it should be possible but you'll need to try it to be sure.

packages/gensx-openai/src/stream.tsx
dereklegenzoff: correct - before we were sending the whole `GSXTool` to the OpenAI method which feels precarious. Now we're just sending the info the API expects

packages/gensx-openai/src/tools.tsx
dereklegenzoff: Long term we might actually want to own this code because we will need to make this work with other providers and they will all have small quirks of different syntax and such i expect.

examples/gsxChatCompletion/index.tsx
dereklegenzoff: Great feedback. I'll take a stab at (1). 

For (2), i thought about this a bit. I briefly changed it to `function` but thought it felt a bit strange because if someone wants to run the tool themselves they would do `const result = await testTool.function({ input: "test" });`

How about either `run` or `runTool`? Or some other verb? CrewAI uses a `_run` method and Llama Stack has you define a `run` method so that seems to be the current standard convention. Langchain uses `invoke`

dereklegenzoff: I particularly want to avoid `function` or `fn` because inside the tool definition itself there's already a `function` which is the json schema

PR #255, opened 2025-02-18
feat: Add `workflow.run` for executing workflows in a fresh context

### LINKED ISSUE #243: Add `gsx.workflow` entrypoint for starting a top level workflow
Currently we have a single API `gsx.execute` method, but we have two scenarios:

1. Starting a workflow within a fresh context
2. Executing a child component that continues the existing context

Currently `gsx.execute` does not allow for (1). If you run subsequent `gsx.execute` calls, you only get checkpoints for the first one. So we need a new method designed for top level use only. 

Proposed API: 

```tsx
const wf = gsx.workflow("Workflow Name", Component);
const result = await wf.run(props);
```

jmoseley:
## Proposed changes

Fixes #243 .

Add a top level `gsx.workflow` that creates a reusable function for executing a workflow:

```
const Component = gsx.Component<{ foo: string }, string>(({ foo }) => {
  return foo;
});

const workflow = gsx.workflow("MyCoolWorkflow", Component);

async main() {
  // These two executions run in separate contexts
  const result = workflow.run({ foo: 'hello' }); // result = 'hello'
  const result2 = workflow.run({ foo: 'bar' }); // result2 = 'bar'
}
```


packages/gensx/tests/execute.test.tsx
```diff
@@ -0,0 +1,58 @@
+import { setTimeout } from "timers/promises";
+
+import { expect, suite, test } from "vitest";
+
+import { gsx, Streamable } from "../src";
+
+suite("execute", () => {
+  const WorkflowComponent = gsx.Component("test", () => "hello");
+
+  suite("workflow", () => {
```
EvanBoyle, line 10: Should we have a test that runs two workflows in parallel and verifies that both get checkpoints?

packages/gensx/src/types.ts
```diff
@@ -42,25 +42,40 @@ export type DeepJSXElement<T> = T extends (infer Item)[]
...
+/**
+ * A component that returns either:
+ * - The output type O directly
+ * - JSX that will resolve to type O
+ * - A promise of either of the above
+ *
+ * Use branding to preserve output type information.
+ * This allows direct access to the output type O while maintaining
+ * compatibility with the more flexible JSX composition system.
+ */
```
EvanBoyle, line 71: nit: maybe move the comment down to be right above the readonly declarations below

packages/gensx/src/execute.ts
```diff
@@ -0,0 +1,58 @@
...
+  const result = (await resolveDeep(element)) as T;
+  context.checkpointManager.write();
+  return result;
+}
+
+type RunResult<P> = P extends { stream: true }
+  ? Promise<Streamable>
+  : Promise<string>;
+
+// Overload for BrandedGsxComponent
```
EvanBoyle, line 26: nit, we eliminated branded and just rolled that into the component definitions.

EvanBoyle (approved):
🚢

packages/gensx/src/checkpoint.ts
```diff
@@ -594,6 +594,11 @@ export class CheckpointManager implements CheckpointWriter {
       // Handle root node case
       if (!this.root) {
         this.root = node;
+
+        // If the workflow name is set, update the root node name.
+        if (this.workflowName) {
+          this.root.componentName = this.workflowName;
```
jmoseley, line 600: @EvanBoyle just FYI. I realized we weren't doing anything with the name parameter, so for now I am just setting the `componentName` on the root node, as this is what we are currently using in the console to extract the name of the workflow. Eventually we need to update the checkpoint API to separate the two concepts.

PR #258, opened 2025-02-18
feat: Improving Visualization for GSXChatCompletion

### LINKED ISSUE #244: Cleanup visualization for GSXChatCompletion and related methods
We recently added a new `GSXChatCompletion` to support structured output, tools, and streaming. The existing `ChatCompletion` was refactored to be composed with `GSXChatCompletion`, and `GSXChatCompletion` is itself composed of a few other helper components. 

Image

While there is upside in the composability, it comes with the downside of some redundancy in the visualization. We should figure out how to clean this up. 

Eventually we might want to offer component options to control what subcomponents get rendered vs hidden by default. But to start we should probably consider a simpler code sharing pattern to hide the bits we don't want to expose. For instance: 

```tsx
const GsxCompletion = gsx.Component("", ()=> {
  // call a function rather than a component
  structuredOutputImpl(props);
}

const StructuredOutput = gsx.Component("", structuredOutputImpl);
```

dereklegenzoff:
Fixes #244 

Introduces implementation functions for:
- ToolExecutor
- ToolCompletion
- StructuredOutput
- StreamCompletion
- GSXChatCompletion (only used in ChatCompletion)

This effectively means that the top level completion (either `ChatCompletion` or `GSXChatCompletion`) will always be shown along with `OpenAIChatCompletion` but no extra middleware is shown. I think this is a good balance because users always get to see the input/output of their top level component and the raw input/output to OpenAI. 

We could also choose to hide `OpenAIChatCompletion` although we do that at the cost of removing some fidelity of what the user gets to see (there is new info displayed when `ChatCompletion` is used and when`GSXChatCompletion` is used with structured outputs)

## Structured Outputs

`StructuredOutput` is no longer shown:

#### After changes

image

#### Before

image

## Tools

`ToolCompletion` and `ToolExecutor` are no longer shown:

#### After changes

image

#### Before

image

## Chat Completion

`GSXChatCompletion` is no longer shown, only when `ChatCompletion` is used:

#### After changes

image

#### Before

image

## Streaming GSXChatCompletion

`StreamComponent` is no longer shown:

#### After changes

image

#### Before

image


packages/gensx-openai/src/gsx-completion.tsx
```diff
@@ -108,7 +122,7 @@ export const ChatCompletion = gsx.StreamComponent<ChatCompletionProps>(
       return generateTokens();
     } else {
       const response = await gsx.execute<ChatCompletionOutput>(
-        <GSXChatCompletion {...props} stream={false} />,
+        gsxChatCompletionImpl({ ...props, stream: false }),
       );
```
dereklegenzoff, line 126: @EvanBoyle, @jmoseley - is this pattern acceptable? it felt strange to use gsx.execute with a function but it works fine and it was giving me issues without the execute

jmoseley: Yes I think this is an ok way to skip the checkpoints for these. Perhaps we need a way to configure `execute` to not record a node?

What issues were you running into?

dereklegenzoff: I see this error if i don't use gsx.execute: 

Unexpected `await` of a non-Promise (non-"Thenable") value.eslint[@typescript-eslint/await-thenable](https://typescript-eslint.io/rules/await-thenable)

Perhaps it's really just a linting thing though? i never tried to run it

jmoseley: oh yeah that is a linter issue. But using execute is ok.

dereklegenzoff: ahh i guess that's because i was using await with `gsxChatCompletionImpl` but it's not async.  but then when i remove the await and run it i do get a vague type error `TypeError: Cannot read properties of undefined (reading '0')`

dereklegenzoff: > oh yeah that is a linter issue. But using execute is ok.

Cool i'll keep it as is for now then

EvanBoyle (approved):
🚢

PR #308, opened 2025-02-20
feat: Handle errors when executing workflow

jmoseley:
## Proposed changes

Add better handling for errors that are thrown when executing a workflow. The `gsx.Workflow` handler now ensures that checkpoints are still sent if the workflow errors, and the error is serialized into a consumable format when sent as part of the checkpoint.


danfhernandez:
I like the simplicity of having it just be an error to start within the metadata 👍

PR #309, opened 2025-02-20
docs: adding prototype agent

dereklegenzoff:


EvanBoyle:
Might be worth looking at https://docs.anthropic.com/en/docs/build-with-claude/tool-use/text-editor-tool as well

PR #313, opened 2025-02-21
fix: clone objects in checkpoints to prevent mutated objects from mes…

### LINKED ISSUE #312: When running a multi-step tool workflow, every OpenAIChatcompletion shows the entire message inputs
This can be repro'd by running the `multiStepTools` or `tools` example in the `gsxChatCompletion` example folder.

Whenever tools are used in a non-streaming context, the checkpoint adds all messages from the entire conversation to every example of `OpenAIChatCompletion` under the `GSXChatCompletion`. This issue doesn't happen with streaming.

In this screenshot, the OpenAIChatCompletion above should have fewer messages than the one below: 
Image

dereklegenzoff:
fixes #312 

The issue was that the `messages` array is getting mutated throughout the workflow when we use tools so we need to create a clone of it to prevent the checkpoint from always rendering the final version of the mutated object.

### DIFF
--- a/packages/gensx/src/checkpoint.ts
+++ b/packages/gensx/src/checkpoint.ts
@@ -4,6 +4,7 @@ import { gzip } from "node:zlib";
 
 import { ComponentOpts, STREAMING_PLACEHOLDER } from "./component";
 import { readConfig } from "./config";
+import { ExecutionContext } from "./context";
 
 const gzipAsync = promisify(gzip);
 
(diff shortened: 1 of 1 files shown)


packages/gensx/src/checkpoint.ts
```diff
@@ -570,6 +571,32 @@ export class CheckpointManager implements CheckpointWriter {
...
+
+    // Handle primitive values
+    if (typeof value !== "object") return value;
+
+    // Handle arrays
+    if (Array.isArray(value)) {
+      return value.map((item) => this.cloneValue(item));
+    }
+
+    // Handle objects that shouldn't be cloned
```
jmoseley, line 589: It feels like there should be more things in this list, but I can't think what.

jmoseley (approved):

packages/gensx/src/checkpoint.ts
dereklegenzoff: Do you think we need to also avoid cloning other framework specific objects like `GSXComponent`s and `JSXElements`? I fed your feedback into cursor and that's what it came up with. Although the hackernewsanalyzer example works fine as is so maybe it's not necessary.

EvanBoyle: Checkpoint lifecycle is roughly: 

1. Add a node to the graph (resolved props only)
2. Update the graph with the completed node when a component finishes executing

Both could trigger checkpoint writes, but when (2) runs all of the outputs are resolved, so we shouldn't see unresolved components in the output. 

Even if someone was passing along a component in props, then the logic that prevents functions from being copied would handle this.

dereklegenzoff: gotcha - so you think this PR is good to go?

PR #317, opened 2025-02-21
docs: self-modifying code demo

EvanBoyle:
WIP

This example demonstrates a self-modifying code agent. The agent:

1. Creates a copy of it's source code
2. Reads context about previous operations
3. Decides on a goal state
4. Modifies it's source code to match the goal state
5. Runs tests to verify the changes
6. Starts a new version of itself to continue the work
7. Spins itself down and allows the new agent to take control

The workflow updates intermediate state on goals, and history of changes that it tried to apply, what succeeded, what failed etc. This log is regularly pushed to the git branch.


examples/self-modifying-code/.gitignore
```diff
@@ -0,0 +1,2 @@
+hn_analysis_report.md
+hn_analysis_tweet.txt
```
jmoseley, line 2:
```suggestion
```

examples/self-modifying-code/context.ts
```diff
@@ -0,0 +1,101 @@
...
+function deserializeContext(json: string): AgentContext {
+  return JSON.parse(json, (key, value) => {
+    if (key === "timestamp" && typeof value === "string") {
+      return new Date(value);
+    }
+    return value;
+  });
+}
+
+export async function readContext(contextFile: string): Promise<AgentContext> {
```
jmoseley, line 40: You might be able to simplify all this to just be `import context from 'context.json;`. Since it modifies the source, when the new process starts up it will get the updated version of the context.

EvanBoyle:
Closing as these changes were merged in https://github.com/gensx-inc/gensx/pull/357

PR #337, opened 2025-02-24
fix: making structured outputs work with tools

### LINKED ISSUE #326: bug: GSXChatCompletion fails when structured output and tools are used together
Fails without error. In addition to a fix we should add a test case to verify.

dereklegenzoff:
Fixes #326

Specifically:
- Adds support for structured outputs to work with tools
- Adds a test for structured output with test
- Removes a couple of unnecessary params from toolExcecutor
- Fixes a couple of places where `ToolExecutor` was still showing in the trace.
- fixes some typos in examples
- Adds a structured output example with tool calls
- remove's `toolExecutorImpl`'s dependency on context since all it was doing was checking for context again

### DIFF
--- a/packages/gensx-openai/src/structured-output.tsx
+++ b/packages/gensx-openai/src/structured-output.tsx
@@ -8,8 +8,8 @@ import {
 } from "openai/resources/chat/completions";
 import { z } from "zod";
 
-import { OpenAIChatCompletion } from "./openai.js";
-import { GSXTool, ToolExecutor } from "./tools.js";
+import { OpenAIChatCompletion, OpenAIContext } from "./openai.js";
+import { GSXTool, toolExecutorImpl } from "./tools.js";
 
 // Updated type to include retry options
 type StructuredOutputProps<O = unknown> = Omit<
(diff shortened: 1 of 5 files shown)


packages/gensx-openai/src/structured-output.tsx
```diff
@@ -49,29 +49,37 @@ export const structuredOutputImpl = async <T,>(
...
       const toolCalls = completion.choices[0].message.tool_calls;
       // If we have tool calls, execute them and make another completion
-      if (toolCalls?.length && tools) {
-        const toolResult = await gsx.execute<ChatCompletionOutput>(
-          <ToolExecutor
-            tools={tools}
-            toolCalls={toolCalls}
-            messages={[...messages, completion.choices[0].message]}
-            model={rest.model}
+      while (toolCalls?.length && tools) {
```
jmoseley, line 63: Is this supposed to be a while loop? I don't see where `toolCalls` is modified.

dereklegenzoff: yeah, it needs to be a while loop in case the model does multiple rounds of tool calls...but you make a good point. This will always only ever run once because there's a return statement in the loop.

dereklegenzoff: okay fixed and confirmed it works by temproarily disabling parallel tool calls: https://app.gensx.com/gensx/executions/01JMX7GP1J46Z8EZE2BSBZ69Q0?workflowName=ToolsWithStructuredOutputWorkflow

jmoseley (approved):

PR #347, opened 2025-02-25
feat: adding @gensx/anthropic

dereklegenzoff:
Creates the `@gensx/anthropic` package modeled after the `@gensx/openai` package.

A few things to call out:
- We use tools for structured outputs. When an output schema is provided, a tool is added and the model is required to call a tool. Tool calls with structured outputs is supported.
- Streaming + tools is not currently supported
- Adds an example project showing the anthropic package in action, equivalent to the `gsxChatCompletion` example


examples/anthropic/.gitignore
```diff
@@ -0,0 +1,2 @@
+hn_analysis_report.md
+hn_analysis_tweet.txt
```
jmoseley, line 2:
```suggestion
```

packages/gensx-anthropic/package.json
```diff
@@ -0,0 +1,57 @@
...
+  ],
+  "author": "GenSX Team",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "@anthropic-ai/sdk": "^0.37.0"
+  },
+  "devDependencies": {
+    "@types/node": "^20.17.11",
+    "@vitest/coverage-istanbul": "^2.1.8",
+    "gensx": "workspace:*",
```
jmoseley, line 43:
```suggestion
```

packages/gensx-anthropic/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "@anthropic-ai/sdk": "^0.37.0",
+    "vite": "^6.1.0",
+    "vite-plugin-dts": "^4.5.0",
+    "vitest": "^2.1.8"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "gensx": "workspace:>=0.2.1",
```
jmoseley, line 53:
```suggestion
    "gensx": "workspace:*",
```

packages/gensx-anthropic/src/tools.tsx
```diff
@@ -0,0 +1,198 @@
...
+interface GSXToolParams<TSchema extends z.ZodObject<z.ZodRawShape>> {
+  name: string;
+  description: string;
+  schema: TSchema;
+  run: (args: z.infer<TSchema>) => Promise<unknown>;
+  options?: {};
+}
+
+// Wrapper for tool parameter schemas
+export class GSXTool<TSchema extends z.ZodObject<z.ZodRawShape>> {
```
jmoseley, line 31: I wonder if there will end up being an issue with having both this type and the "same" type in @gensx/openai. It will be confusing for the developer to create all their tools but not be able to interchange them.

examples/anthropic/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Multi-step tool workflows
+
+## Usage
+
+```bash
+# Install dependencies
+pnpm install
+
+# Set your OpenAI API key
+export OPENAI_API_KEY=<your_api_key>
```
jmoseley, line 18: Anthorpic?

packages/gensx-anthropic/README.md
```diff
@@ -0,0 +1,61 @@
...
+npm install @gensx/anthropic
+```
+
+### Requires `gensx`
+
+This package requires `gensx` to be installed as a peer dependency.
+
+```bash
+npm install gensx
+```
```
jmoseley, line 17:
```suggestion
```

It's not a peer dependency.

packages/gensx-anthropic/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@gensx/anthropic",
+  "version": "0.1.7",
+  "description": "Anthropic integration for GenSX",
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "engines": {
+    "node": ">=18.0.0"
```
jmoseley, line 9:
```suggestion
    "node": ">=18.0.0",
    "pnpm": ">=9.0.0"
```

We need to go update this everywhere.

packages/gensx-anthropic/src/stream.tsx
```diff
@@ -0,0 +1,87 @@
...
+        tools={tools.map((t) => t.definition)}
+        stream={false}
+      />,
+    );
+
+    // If no tool calls, proceed with streaming the original response
+    if (completion.stop_reason !== "tool_use") {
+      return gsx.execute<Stream<RawMessageStreamEvent>>(
+        <AnthropicChatCompletion {...rest} stream={true} />,
+      );
```
jmoseley, line 46: You can use Component.run now (might need to rebase).

```suggestion
      return AnthropicChatCompletion.run({
        ...rest,
        stream: true
      });
```

packages/gensx-anthropic/README.md
dereklegenzoff: should I remove this from the OpenAI README too? This is all copied and pasted with minor updates

packages/gensx-anthropic/src/tools.tsx
dereklegenzoff: Yeah i'm worried about that too. Pretty sure the code is 100% identical. @EvanBoyle and i talked about evantually having a shared package for all this stuff but i'm assuming that can wait?

examples/anthropic/README.md
dereklegenzoff: good catch!

packages/gensx-anthropic/package.json
dereklegenzoff: updated. want me to go ahead and update this on all the other packages?

examples/anthropic/.gitignore
dereklegenzoff: good catch

packages/gensx-anthropic/package.json
dereklegenzoff: good catch, updated

packages/gensx-anthropic/src/tools.tsx
dereklegenzoff: oh nvm, the code is in fact different (had to update the tool type).

Should we consider naming this different? Or just expect folks to import it {GSXTool as AnthropicGSXTool}?

EvanBoyle: I'm personally ok with having users rename the imports when they want to use multiple providers in the same program. I assume that won't happen super frequently. Though open for discussion.

packages/gensx-anthropic/README.md
jmoseley: Yeah probably.

packages/gensx-anthropic/package.json
jmoseley: sure if you feel like it. Not high priority though

jmoseley (approved):

PR #349, opened 2025-02-25
Basic MCP example

danfhernandez:
This example leverages the MCP client sdk to pull down and interact with an MCP server. It converts MCP tools into the appropriate format expected by the AI SDK from vercel as well. 

Unfortunately there's seems to be an issue with client not being properly passed down to children. When running this example, callTool() fails when generate text is called.

### DIFF
--- a/examples/mcp/src/index.tsx
+++ b/examples/mcp/src/index.tsx
@@ -1,0 +1,120 @@
+import { gsx } from "gensx";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StdioClientTransport } from "@modelcontextprotocol/sdk/client/stdio.js";
+import { GenerateText } from "@gensx/vercel-ai-sdk";
+import { openai } from "@ai-sdk/openai";
+import { tool, ToolSet } from "ai";
+import { z } from "zod";
+
+
+interface MCPServerResult {
+  client: Client;
+  availableTools: any;
+}
+interface MCPServerProps {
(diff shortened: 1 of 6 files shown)


examples/mcp/src/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+  availableTools: any;
+}
+interface MCPServerProps {
+  clientName: string;
+  clientVersion: string;
+  serverCommand: string;
+  serverArgs: string[];
+}
+
+const MCPServer = gsx.Component<MCPServerProps, MCPServerResult>(
```
jmoseley, line 21: I think this should be a Provider, and makes the tools available in a context.

danfhernandez: I'd be curious to see what that code looks like. Haven't created a provider yet. 

In terms of provider vs component, I wonder what the rule of thumb should be. I could go either way here. It would be the client you would be passing down which has a set of functions on it to take actions (which actually doesn't work outside of the component where's it defined)

EvanBoyle: I agree with Jeremy in this instance. 

You don't want to have to wrap every ChatCompletion component with an MCPServer component in a real workflow. Makes much more sense to do it at the top of the workflow and then access the tools with `useContext` in any of the subcalls within a tree where you need them.

jmoseley:
Merged via https://github.com/gensx-inc/gensx/pull/358

PR #358, opened 2025-02-26
feat: @gensx/mcp

jmoseley:
## Proposed changes

Add a new package, `@gensx/mcp`, that enables the creation of MCP contexts with a provider and a `useContext` helper. This provider manages the lifecycle of the server (spinning it up and shutting it down).

One thing that was required to make this work that I'm not super stoked about: Needed to add an `onComplete` to the `Provider` interface, so that we can perform tasks _after_ the children of a provider have full resolved. This is necessary for us to close the connection (aka shutdown) the MCP server once that part of the workflow has finished.


EvanBoyle (commented):
> One thing that was required to make this work that I'm not super stoked about: Needed to add an onComplete to the Provider interface, so that we can perform tasks after the children of a provider have full resolved. This is necessary for us to close the connection (aka shutdown) the MCP server once that part of the workflow has finished.

Can we handle the lifecycle via node.js process hooks? This thing is a singleton across the node process, so can we just add a `process.on(exit, ...)` handler? (maybe other hooks/handler we need to attach to as well...).

packages/gensx-mcp/README.md
```diff
@@ -0,0 +1,132 @@
...
+### Using MCP Tools with LLM Providers
+
+You can use MCP tools with any LLM provider in GenSX. Here's an example with OpenAI:
+
+```tsx
+import { createMCPServerContext, MCPTool } from "@gensx/mcp";
+import { GSXChatCompletion, OpenAIProvider } from "@gensx/openai";
+import { gsx } from "gensx";
+
+// Helper function to map MCP tools to GSX tools
```
danfhernandez, line 42: Should this be in the package itself?

jmoseley: Good question. Perhaps it should be in the `@gensx/openai` and `@gensx/anthropic` packages?

They both have a `GSXTool`, but they are different types.

jmoseley: Ok I solved this by enabling both `@gensx/openai` and `@gensx/anthropic` to accept `GSXToolParams` as well as `GSXTool` for the tools parameter, and moved the type of `GSXToolParams` into `gensx` so it can be shared everywhere.

This allows me to add a `asGsxTool()` helper to the `MCPTool` wrapper, simplifying the logic, but not needing to introduce dependencies between `@gensx/mcp`, `@gensx/openai`, or `@gensx/anthropic`.

jmoseley: cc @dereklegenzoff

jmoseley:
@EvanBoyle I tidied this up, but opted to keep the `onComplete` functionality. However, I did add support for passing a pre-connected `Client`, so that the developer has the option to manage the lifecycle themselves if they like.

packages/gensx-mcp/README.md
dereklegenzoff: sounds great

PR #361, opened 2025-02-27
docs: adding/updating docs for all llm providers

dereklegenzoff:
- Updates OpenAI docs with latest components
- Adds anthropic docs
- Adds vercel ai sdk docs


danfhernandez (approved):

website/docs/src/content/component-reference/vercel-ai.mdx
```diff
@@ -0,0 +1,334 @@
...
+console.log(result.url); // URL to the generated image
+```
+
+##### Props
+
+The `GenerateImage` component accepts all parameters from the Vercel AI SDK's `experimental_generateImage` function:
+
+- `prompt` (required): The text description of the image to generate
+- `model` (required): The image generation model to use (from Vercel AI SDK)
+- Plus all other parameters supported by the Vercel AI SDK
```
EvanBoyle, line 299: Nit: I believe this should be "plus any other optional parameters..."

The goal here should be to document required parameters, and link to the vercel docs for other supported parameters and output types.

EvanBoyle (approved):
Overall, looks great. 

I also feel like we should link to the reference docs for the provider in question (direct links to the props type and return type) so that it is easy for users to figure out what parameters exist. What do you think?

PR #368, opened 2025-02-27
docs: Salty ocean example

danfhernandez:
Example that demonstrates calling the same prompt across models.


examples/salty-ocean/README.md
```diff
@@ -0,0 +1,53 @@
+# Why is the Ocean Salty? - GenSX Example
```
dereklegenzoff, line 1: shouldn't we call this "model comparison" or something like that? In a year it seems like people will be confused why there's an example called salty ocean and what it does

examples/salty-ocean/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "my-gensx-app",
```
dereklegenzoff, line 2: This should be something like `@gensx-examples/salty-ocean` or `@gensx-examples/model-comparison`

examples/salty-ocean/README.md
danfhernandez: Updated!

examples/salty-ocean/package.json
danfhernandez: Updated!

danfhernandez:
> Agree with Derek, calling this something like "/model-comparison" or "salty-ocean-model-comparison" would probably make it easier to discover.


Totally, makes sense. I updated it so that the naming is more generic

> Not a blocker, but the only other comment is that for the future the more models an individual example uses the fewer people will be able to run it. Being able to set creds for either groq/openai or both might be a pattern to take advantage of in the future, but feel free to just ship it and think about this for the next one.
Yeah I'll update this on a follow on pr. Def agree

Issue #227, opened 2025-02-07
Breakpoints and stack traces for debugging examples


jmoseley:
It would be useful to improve the project linking in the monorepo to enable stack traces and breakpoints that cross the boundary from the examples to the gensx packages. This would improve the developer experience and ease troubleshooting in the future.

Right now:
* Stack traces point to bundled code, making it challenging to trace the flow through GenSX
* You cannot set breakpoints in the gensx package and have the example pause

Ideally:
* Node debugger integration with VS code launch tasks for the examples
* Breakpoints in the gensx and @gensx/openai package will work 
* Stack traces will show the right lines from the packages

Also Nice to have:
* Ensure that sourcemaps are configured properly so that projects that depend on GenSX also get high quality stack traces

PR #371, opened 2025-02-28
feat: Don't minify packages

### LINKED ISSUE #323: Do not minify code, and ensure source maps work when debugging
Semi related to https://github.com/gensx-inc/gensx/issues/227

jmoseley:
## Proposed changes

* Build using `tsc` instead of vite, so that packages are not minified and sourcemap resolution works well.
* Simplify shared tsconfigs
* Add launcher for examples so that we can use breakpoints etc.
* Add vitest workspace

Fixes #323 and #227


EvanBoyle (approved):
🚢

PR #372, opened 2025-02-28
feat: improving the checkpoint output for tools and structured outputs

### LINKED ISSUE #221: Customize the way that tools are rendered in checkpoints (custom toJSON to show the openAPI schema, remove all of the zod native functions)

dereklegenzoff:
Fixes #221 

- Small fix to checkpointing logic to us toJSON() to copy objects where possible
- Adds a toJSON() method to GSXTool
- Extends zodSchema to implement a toJSON() method


packages/gensx/src/checkpoint.ts
```diff
@@ -591,6 +591,12 @@ export class CheckpointManager implements CheckpointWriter {
     if (Symbol.asyncIterator in value) return value;
     if (ArrayBuffer.isView(value)) return value;
 
+    // Check for toJSON method before doing regular object cloning
+    const objValue = value as { toJSON?: () => unknown };
+    if (typeof objValue.toJSON === "function") {
+      return this.cloneValue(objValue.toJSON());
+    }
```
EvanBoyle, line 598: Can we add tests to cover this?

EvanBoyle (commented):
Just needs tests but otherwise I'm happy with this approach.

packages/gensx/src/checkpoint.ts
dereklegenzoff: fair, i'll add a test

PR #395, opened 2025-03-04
docs: Add open router example

danfhernandez:
Added example showing how to use OpenRouter. It's a simple example that uses the OpenAI provider and a simple option through the provider prop.

### DIFF
--- a/website/docs/src/content/component-reference/openrouter.mdx
+++ b/website/docs/src/content/component-reference/openrouter.mdx
@@ -3,7 +3,7 @@ title: OpenRouter
 description: Using GenSX with OpenRouter
 ---
 
-# OpenRouter Integration
+# OpenRouter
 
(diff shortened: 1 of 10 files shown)


EvanBoyle (commented):
@danfhernandez looks great, but can we add (1) a light-weight /examples page, and (2) a page for OpenRouter under component reference? 

For (2), it can just be a few lines: 

OpenRouter can be used via the OpenAI provider (add link to docs): 

```tsx
code example...
```

website/docs/src/content/component-reference/openrouter.mdx
```diff
@@ -0,0 +1,117 @@
+---
+title: OpenRouter
+description: Using GenSX with OpenRouter
+---
+
+# OpenRouter
+
+[OpenRouter](https://openrouter.ai) provides a unified API to access various AI models from different providers. You can use GenSX with OpenRouter by configuring the OpenAIProvider component with OpenRouter's API endpoint.
```
EvanBoyle, line 8: We should link to OpenAI component docs here.

EvanBoyle (approved):

PR #440, opened 2025-03-07
feat: support structured outputs via tools in the OpenAIProvider

### LINKED ISSUE #348: Support Structured outputs in @gensx/openai with tools
To make structured outputs work with other openai-compatible APIs, we should add a way to support structured outputs via tool calls. 

We need to consider if we will switch over to this for all models (including openai models) or have a fork in the logic.

dereklegenzoff:
fixes #348 

This PR updates the OpenAIProvider to make it possible to use structured outputs with other OpenAI Compatible APIs. If the base url is from openai or azure openai, it will use `response_format` to create the structured output, otherwise it will use tools.

This PR also adds a new prop `structuredOutputStrategy`  to allow users to control the method used for structured outputs. By default, the param is set to `default` but users can also specify `tools` or `response_format`.


packages/gensx-openai/__mocks__/openai.ts
```diff
@@ -81,6 +81,14 @@ const mockCreateMethod = vi
     }
     // Handle tool calls
     else if (params.tools && params.tools.length > 0) {
+      // Check if there's an output_schema tool in the tools array
+      const hasOutputSchemaTool = params.tools.some(
+        (tool) =>
+          typeof tool === "object" &&
+          "function" in tool &&
+          tool.function.name === "output_schema",
```
jmoseley, line 89: Should we namespace this to reduce the slim chance of collisions?

```suggestion
          tool.function.name === "gensx_output_schema",
```

packages/gensx-openai/src/structured-output.tsx
```diff
@@ -29,70 +32,153 @@ type StructuredOutputProps<O = unknown> = Omit<
...
+): GSXTool<any> => {
+  // Create a wrapper schema that will contain the output
+  const wrapperSchema = z.object({
+    output: outputSchema,
+  });
+
+  return GSXTool.create({
+    name: "output_schema",
+    description:
+      "When producing a final output, always call this tool to produce a structured output",
```
jmoseley, line 47: Do you think this description is specific enough? It feels a bit light on details and commands. But I dunno, sometimes less is more for these things.

packages/gensx-openai/src/structured-output.tsx
```diff
@@ -29,70 +32,153 @@ type StructuredOutputProps<O = unknown> = Omit<
...
-  const tools = toolsParams?.map((t) =>
-    t instanceof GSXTool ? t : new GSXTool(t),
-  );
+  // Get the OpenAI client from context
+  const context = gensx.useContext(OpenAIContext);
+  if (!context.client) {
+    throw new Error(
+      "OpenAI client not found in context. Please wrap your component with OpenAIProvider.",
+    );
+  }
```
jmoseley, line 78: @danfhernandez fixed it so we don't need this anymore.

```suggestion
```

packages/gensx-openai/src/structured-output.tsx
```diff
@@ -29,70 +32,153 @@ type StructuredOutputProps<O = unknown> = Omit<
...
+
+  // Check if the baseURL is for OpenAI or Azure OpenAI
+  const baseURL = context.client.baseURL;
+  const isOpenAIModel =
+    !baseURL ||
+    baseURL.includes("openai.com") ||
+    baseURL.includes("openai.azure.com");
```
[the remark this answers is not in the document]
jmoseley: Can we look at the model name as well or instead?

jmoseley (approved):
I think this needs a bit deeper testing, otherwise LGTM.

packages/gensx-openai/src/structured-output.tsx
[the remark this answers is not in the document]
dereklegenzoff: Yeah - if they're using openrouter, they'd have to use the `structuredOutputStrategy` prop.

We could look at model name but then we'll be constantly updating the list and for newer models it won't work unless users upgrade so that feels worse.

packages/gensx-openai/src/structured-output.tsx
jmoseley: I guess I remember trying to get this to work effectively and it was problematic, depending on the model. 

That being said, here is what I settled on 😆 https://github.com/cortexclick/cortex/blob/a9cee41916697b0a6a8c176c3d957ff81178ece5/cortex-api/content/helpers/createContentExecutors/langgraph/utils/structuredReactAgent.ts#L40

packages/gensx-openai/__mocks__/openai.ts
dereklegenzoff: hmm interesting idea. I like it in principle but the model has no concept of what gensx is so i could see this hurting performance. in my mind `output_schema` as the tool name is what makes its purpose super obvious to the model and the description is less important.  

I'm a bit less worried about a name collison here than i am with the `outputSchema` prop we added

packages/gensx-openai/src/structured-output.tsx
dereklegenzoff: removing this, thanks!

packages/gensx-openai/__mocks__/openai.ts
jmoseley: 👍

jmoseley (approved):

PR #447, opened 2025-03-11
docs: Fix up example packages and docs for running examples.

jmoseley:
## Proposed changes

* Fix up the build configs and package names for examples
* Add command for running example from the root of the repo, that includes the build step if necessary
* Update docs.


examples/README.md
```diff
@@ -4,20 +4,10 @@ This folder contains a number of different examples to help you get up and runni
...
-pnpm build:all
-```
-
-From there, follow the instructions in the README of the example you want to run.
-
-Alternatively, you can run the examples directly from the root directory of the repo using the following command:
-
-```bash
-OPENAI_API_KEY=<my api key> turbo run start --filter="./examples/blogWriter"
+OPENAI_API_KEY=<my api key> pnpm start:example <example-name>
```
dereklegenzoff, line 10: would this still work even if a user hasn't run pnpm install?

examples/README.md
```diff
@@ -4,20 +4,10 @@ This folder contains a number of different examples to help you get up and runni
 
 ## Running the Examples
 
-To run the examples, start by installing and building everything from the root directory of the repo. This will install the dependencies and build all of the packages and examples
+From the root of the repo, run the following command to build packages and run an example:
```
dereklegenzoff, line 7: I wonder if we should still document how to run an example from within the folder. We have the instructions in all the READMEs already so maybe we just have a sentence somewhere that says "You can also build and run an example from it's folder. Check the example's README for details."

examples/all-models-metadata/src/index.tsx
```diff
@@ -24,13 +24,13 @@ interface OpenRouterModel {
...
   };
-  per_request_limits: any | null;
+  per_request_limits: unknown;
 }
 
 // This interface is more generic to accommodate both OpenAI and OpenRouter responses
 interface ListModelsOutput {
   models: {
-    data: any[]; // Using any[] to accommodate different model structures
+    data: unknown[]; // Using any[] to accommodate different model structures
```
dereklegenzoff, line 33: nit: the comment is now outdated
```suggestion
    data: unknown[]; 
```

dereklegenzoff (approved):
looks good, left a few small comments.

examples/README.md
jmoseley: yes, it uses turbo and turbo is aware of the dependencies.

examples/README.md
jmoseley: good idea

examples/README.md
dereklegenzoff: perfect!

PR #457, opened 2025-03-12
fix: Use @types/node for minimum supported version (v18) in packages.

jmoseley:
## Proposed changes

Ensure that we are using the right types for node. We support v18 with `@gensx/*`, so don't build with anything else.


pnpm-workspace.yaml
```diff
@@ -19,3 +18,9 @@ catalog:
   glob: "^10.3.10"
   gray-matter: "^4.0.3"
   "@types/glob": "^8.1.0"
+
+catalogs:
+  examples:
+    "@types/node": "^22"
```
EvanBoyle, line 24: Why v22 for examples?

EvanBoyle (approved):

pnpm-workspace.yaml
EvanBoyle: If this is necessary, we should document it somewhere.

jmoseley: I chose 22 as it is the current active version.

PR #461, opened 2025-03-13
feat: Beta commands for serverless execution

jmoseley:
## Proposed changes

* `build`, `deploy`, and `run` commands for interacting with the serverless execution platform.
* Do not make these commands available unless you are part of the `gensx` org.


packages/gensx/src/commands/new.ts
```diff
@@ -227,6 +229,21 @@ export async function newProject(
         spinner.succeed();
       }
 
+      // Create gensx.yaml configuration file with project name (required)
+      spinner.start("Creating project configuration file");
+      const projectName = path.basename(absoluteProjectPath);
```
EvanBoyle, line 234: TODO: we should prompt users to supply this eventually.

EvanBoyle (approved):
Overall looks like a solid sketch. I will have more feedback on CLI shape plus params once I get my hands on it for dogfooding. 

We will very quickly want to move to a model where we have a `dev` command as well so everything from building/deploying/running in the cloud/developing locally is handled by the CLI.

PR #465, opened 2025-03-14
docs: documenting component.run

dereklegenzoff:
- Adds light docs on component.run
- Updates reflection example and page to use component.run
- Updates reference docs to not use gensx.execute

### DIFF
--- a/website/docs/src/content/basic-concepts.mdx
+++ b/website/docs/src/content/basic-concepts.mdx
@@ -134,13 +134,17 @@ Rather than having to run an entire workflow to test a change to a single compon
 
 ## Executing sub-workflows
 
-In some cases, you may need to run a sub-workflow within a larger workflow. You can do this by using `gensx.execute()`:
+In some cases, you may need to run a sub-workflow within a larger workflow. You can do this using both `Component.run()` and `gensx.execute()`:
 
 ```tsx
+// Using Component.run()
+const result = await MyComponent.run({ input: "some data" });
+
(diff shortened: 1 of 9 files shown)


website/docs/src/content/basic-concepts.mdx
```diff
@@ -134,13 +134,17 @@ Rather than having to run an entire workflow to test a change to a single compon
 
 ## Executing sub-workflows
 
-In some cases, you may need to run a sub-workflow within a larger workflow. You can do this by using `gensx.execute()`:
+In some cases, you may need to run a sub-workflow within a larger workflow. You can do this using both `Component.run()` and `gensx.execute()`:
```
jmoseley, line 137: I think we should remove the reference to `gensx.execute`. I think `Component.run` covers most of the use cases.

The only use case that it enables is executing a tree without assigning to a single component:

```
const result = gensx.execute(<Parent><Child /></Parent>);
```

Maybe thats worth keeping it, I'm not sure.

dereklegenzoff: fair point. i'm a bit on the fence; I could go either way. What do you think @EvanBoyle and @danfhernandez ?

EvanBoyle: I think it is worth keeping somewhere though I wouldn't lead front and center. 

We need a page that discusses workflows vs components and how to organize code at some point.

PR #466, opened 2025-03-15
docs: adding patterns docs for tools

dereklegenzoff:
Adds a simple patterns docs showing how to use tools in GenSX.

### DIFF
--- a/website/docs/src/content/patterns/tools.mdx
+++ b/website/docs/src/content/patterns/tools.mdx
@@ -7,11 +7,11 @@ description: Using tools with GenSX
 
 Using tools provides a powerful and flexible way to extend the capabilities of an LLM. GenSX offers several different ways for you to work with tools:
 
-- Using the `GSXChatCompletion` component from `@gensx/openai` or `@gensx/anthropic` for fully automated tool calling
-- Using the `OpenAIChatCompletion` or `AnthropicChatCompletion` components from `@gensx/openai` or `@gensx/anthropic` and following the same pattern as described in the [OpenAI](https://platform.openai.com/docs/guides/function-calling?api-mode=responses) and [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview) docs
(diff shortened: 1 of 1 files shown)


website/docs/src/content/patterns/tools.mdx
```diff
@@ -0,0 +1,102 @@
...
+  description: "get the weather for a given location",
+  schema: weatherSchema,
+  run: async ({ location }: WeatherParams) => {
+    console.log("getting weather for", location);
+    const weather = ["sunny", "cloudy", "rainy", "snowy"];
+    return Promise.resolve({
+      weather: weather[Math.floor(Math.random() * weather.length)],
+    });
+  },
+});
```
jmoseley, line 51: I tihnk its worth mentioning that this can just be the params too:

```suggestion
const weatherTool: gensx.GSXToolParams = {
  name: "get_weather",
  description: "get the weather for a given location",
  schema: weatherSchema,
  run: async ({ location }: WeatherParams) => {
    console.log("getting weather for", location);
    const weather = ["sunny", "cloudy", "rainy", "snowy"];
    return Promise.resolve({
      weather: weather[Math.floor(Math.random() * weather.length)],
    });
  },
}
```

jmoseley (approved):

website/docs/src/content/patterns/tools.mdx
EvanBoyle: I think I missed that we changed this to ToolParams at some point. Should it be ToolProps, to be more consistent with naming conventions for the rest of the programming model ?

dereklegenzoff: Here's a PR for the rename assuming we decide to move forward with it: https://github.com/gensx-inc/gensx/pull/469

dereklegenzoff: also, added a note in the doc regarding Jeremy's original point

PR #467, opened 2025-03-15
feat: Add @gensx/cursor-rules package for automatic cursor rule updates

### LINKED ISSUE #450: Add CLI command for updating cursor rules
We should add a CLI command that allows you to update the cursor rules in your project. Currently when we push updates to the rules, the only things users can do is copy and paste the new rules which is a pain.

We should make sure the command updates our cursor rules but doesn't interfere with other cursor rules files.

jmoseley:
## Summary
- Adds a new  package that provides a way to automatically update cursor rules in GenSX projects
- The package includes a postinstall script that copies the cursor rules to the project's  directory
- It handles both new installations and updates, preserving any custom rules users might have added

Closes #450


EvanBoyle (approved):
Looks good to me. I guess the only downside here is that if the user changed the gensx cursor rules, they would lose those edits. But they will see it in the diff, so probably fine.

jmoseley:
Yeah I figured there was a tradeoff, and I made the assumption that people probably won't edit our rules, but add their own.

However, for the CLAUDE.md package, that assumption doesn't hold, so I opted for a comment delimited section with a warning that anything inside of it will be replaced during updates. We could do something similar here probably.

PR #471, opened 2025-03-15
feat: add @gensx/windsurf-rules package

jmoseley:
## Summary
- Adds a new @gensx/windsurf-rules package that automatically installs a .windsurfrules file to GenSX projects
- Helps Windsurf's Cascade AI provide project-specific guidance for working with GenSX
- Similar to the cursor-rules and claude-md packages previously implemented


EvanBoyle (approved):
Looks a little light compared to cursorrules, but a good start! 

What a pain in the ass that all of these tools require different files in slightly different formats. 

At some point we should consider generating all of this stuff from a single source of truth. Perhaps file an issue and we can worry about this stuff another day.

jmoseley:
Yeah I agree generating these will be useful. It's going to be tedious updating them all.

jmoseley:
I bet we will see some convergence at some point. I noticed Claude code, for example, does check the cursor rules folder sometimes.

jmoseley:
I'll update it to reflect the existing content, don't know what went missing there.

jmoseley:
#vibecoding

PR #481, opened 2025-03-17
fix: Fix schema extraction for structured types

jmoseley:
* Fix cli binary build and add test

Fix resolution of named Props and structured output when bundling a workflow and producing the schema. There was a bug that was returning `{ "type": "object" }` for any output type that was not a `string` or `Streamable`, and there was an issue resolving named types for props. This resolved both of those problems.

### DIFF
--- a/.github/workflows/release-cli-binary.yml
+++ b/.github/workflows/release-cli-binary.yml
@@ -32,15 +32,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [win, linux, macos-x64, macos-arm64]
+        target: [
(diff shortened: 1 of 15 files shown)


EvanBoyle (approved):

EvanBoyle (approved):
Looks good but we should set up tests for schema conversion ASAP. 

Probably the first layer we'll find bugs and we should lock in behavior as we make changes. Fine to do in a later follow up with everything we have on our plates right now.

jmoseley:
https://github.com/gensx-inc/gensx/issues/482

PR #485, opened 2025-03-18
feat: Convert IDE rules packages to npx commands

jmoseley:
Change the installation model from an installable package to an `npx` command for the AI IDE rules packages.


jmoseley:
Test will unfortunately fail until we publish

danfhernandez (approved):

PR #498, opened 2025-03-19
fix: Improve tests for rules packages.

jmoseley:
* Add tests that do not require a release to validate
* Remove an unnecessary dependency that is causing flakey tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "prettier": "^3.0.0",
     "turbo": "^2.4.4",
     "typescript": "catalog:",
-    "vite": "^6.2.2"
+    "vite": "catalog:"
   },
   "dependencies": {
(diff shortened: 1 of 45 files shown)


EvanBoyle (commented):
Any reason these test files are JS instead of TS?

jmoseley:
> Any reason these test files are JS instead of TS?

Just didn't feel like getting TS running in those packages. I can if you think its important.

EvanBoyle:
Just kinda weird to have arbitrary JS projects in our repo. On one hand it feels like (especially since our project more or less requires TS for schema gen) everything should be TS in this repo out of principal, consistency of tools, workflow etc. On the other hand I don't want you to waste time. So choose whatever you think makes the most sense given all of that.

PR #509, opened 2025-03-20
fix: Fix type inference for StreamComponent and Provider

### LINKED ISSUE #491: Autocomplete for ChatCompletion is broken
Looks like type inference went missing:

Image

jmoseley:
This fixes the type inference for `StreamComponent` and `Provider`, which was trickling down to `OpenAIProvider` and `ChatCompletion`. 

The root cause was that our layout of types was causing the Typescript compiler to emit inline `import("node_modules/...")` statements instead of types, and failing to resolve these. I refactored the types to move most of them into `types.ts`, and that helped. I also added more tests to validate that types are behaving the way we want, and improved the mechanism to disallow extra prompts for StreamComponent, and fixed the type inference for the child type parameter for StreamComponent.

Fixes #491


packages/gensx-core/src/component.ts
```diff
@@ -29,12 +29,15 @@ export type WithComponentOpts<P> = P & {
   componentOpts?: ComponentOpts;
 };
 
+type ComponentProps<P extends object & { length?: never }> =
```
jmoseley, line 32: `ComponentProps`, and `StreamComponentProps` below are the root of the fix.

danfhernandez (approved):

PR #513, opened 2025-03-21
feat: Vector DB example.

jmoseley:
* Add an example that uses a vector DB, and demonstrates a few common things like chunking, generating embeddings, and querying the VectorDB
* Add the `OpenAIEmbeddings` component.


examples/vector-db/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@examples/vector-db",
```
dereklegenzoff, line 2: I feel like vector-db isn't the right name for this since there isn't actually a vector db involved. Maybe we call it "rag" or "retrieval-augemented-generation"

examples/vector-db/README.md
```diff
@@ -0,0 +1,26 @@
+# NuLang Translator
```
dereklegenzoff, line 1: We've typically had this match the example folder name a bit more closely. Maybe we call this "Retrieval Augmented Generation "and can include details in the scenario in the text.

Was also chatting with Evan about this and i wonder if we should just make this RAG over the GenSX docs so it's a more business-y scenario. If you wanted, i'd be happy to put together a scraper side of this so the example could essentially allow a user to paste in a url to their website and get RAG over their website...or we could just leverage the monorepo and point to the gensx docs

EvanBoyle:
I love the example and thank you @jmoseley to put it together. I will mention that I spent more time trying to understand the use case for this than I spent reading the code. It feels a little hard to grok and academic, and I wonder if a more straightforward use case that our users could related to more easily would have a bigger impact here. 

As derek suggested, we could do gensx docs. We could even make the example self-contained (just index the md/mdx content in the docs/website folder) so that we don't require web scraping.

jmoseley:
Good point. I started with that in mind, then I went to build new documentation, and it kind of snowballed from there.

Happy to move this into its own toy example. RAG on our docs seems like a much better choice.

PR #524, opened 2025-03-26
feat: support a local dev server

dereklegenzoff:
Adding a local dev server including the following features:
- Support for all GenSX Workflow APIs with identical API contract
   - List Workflow
   - Get Workflow
   - Run Workflow
   - Start Workflow
   - Get Workflow Execution
   - List Workflow Executions 
- Strong validation/error handling
- Detects and validates schemas on the request
- Supports hot reloading
- Support for Swagger UI
- Tests for the start command and the dev server


packages/gensx/package.json
```diff
@@ -51,12 +51,15 @@
     "@rollup/plugin-json": "^6.0.0",
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-typescript": "^11.1.6",
+    "@gensx/openai": "workspace:*",
```
jmoseley, line 54: hmm. I think this is a circular dependency.

packages/gensx/src/commands/start.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    // 1. Validate file exists and is a TypeScript file
+    const absolutePath = resolve(process.cwd(), file);
+    if (!existsSync(absolutePath)) {
+      throw new Error(`File ${file} does not exist`);
+    }
+
+    if (!file.endsWith(".ts") && !file.endsWith(".tsx")) {
+      throw new Error("Only TypeScript files (.ts or .tsx) are supported");
+    }
```
jmoseley, line 44: Really? Is there a good reason to not support `js` or `jsx`?

packages/gensx/package.json
dereklegenzoff: oh yeah not sure why that's in there; we don't need it.

packages/gensx/src/commands/start.ts
dereklegenzoff: oh i was under the impression that it had to be typescript so users could tell us the input/output schema for the workflows. Happy to relax this though.

packages/gensx/src/utils/bundler.ts
```diff
@@ -101,6 +101,7 @@ export function getRollupConfig(
         jsx: "react-jsx",
         jsxImportSource: "@gensx/core",
         tsconfig: "./tsconfig.json",
+        module: "NodeNext",
```
dereklegenzoff, line 104: @jmoseley - i kept getting errors saying we needed to specify NodeNex here so i added it in; not sure if it's okay....perhaps a moot point since you're changing the bundling anyways

examples/gsxChatCompletion/gensx.yaml
```diff
@@ -0,0 +1 @@
+projectName: derek-sandbox
```
jmoseley, line 1: I think we should get rid of `gensx.yaml` for now, bring it back when we are ready to talk about this stuff.

dereklegenzoff: yep, good call. same with the workflow.tsx. I had removed them at one point but accidentally added them back in  lol

packages/gensx/src/utils/bundler.ts
```diff
@@ -68,9 +86,54 @@ export async function bundleWorkflow(
...
+    // Move to the project directory
+    const originalDir = process.cwd();
+    process.chdir(packageJsonDir);
+
+    // Install @vercel/ncc if not already installed
+    try {
+      await execPromise("npx --no-install @vercel/ncc --version");
+    } catch (_error) {
+      await execPromise("npm install -g @vercel/ncc");
+    }
```
jmoseley, line 115: Why do we need to force it to be installed globally? Does `npx @vercel/ncc` not work?

packages/gensx/src/dev-server.ts
```diff
@@ -0,0 +1,1192 @@
...
+   * Set up server routes
+   */
+  private setupRoutes(): void {
+    // Add middleware
+    this.app.use("*", logger());
+    this.app.use("*", cors());
+
+    // Add project validation middleware
+    this.app.use(
+      `/org/${this.org}/projects/:projectName/*`,
```
jmoseley, line 274: Is this middleware necessary? Looks like all the other routes just hardcode the project name, so I assume this is to return a slightly improved 404 message? If we do think its necessary, should we have a similar one for the org?

packages/gensx/src/dev-server.ts
```diff
@@ -0,0 +1,1192 @@
...
+        "Content-Type": "application/stream",
+        "Transfer-Encoding": "chunked",
+      },
+    });
+  }
+
+  /**
+   * Generate OpenAPI specification dynamically based on server configuration
+   */
+  private generateOpenApiSpec(): Record<string, unknown> {
```
jmoseley, line 640: We should port this over to production soon.

packages/gensx/src/utils/bundler.ts
```diff
@@ -68,9 +86,54 @@ export async function bundleWorkflow(
...
+      `npx @vercel/ncc build ./${relativeWorkflowPath} -o ${distDir} --target es2022`,
+    );
+
+    if (stderr) {
+      // Process stderr if needed
+    }
+
+    // Create tar.gz file (for compatibility with Docker build)
+    process.chdir(distDir);
+    await execPromise(`tar -czvf ../dist.tar.gz *`);
```
jmoseley, line 128: Is this used for anything?

packages/gensx/src/utils/bundler.ts
```diff
@@ -68,9 +86,54 @@ export async function bundleWorkflow(
...
   } catch (error) {
-    console.error("Error bundling workflow", error);
+    console.error("Error bundling workflow with Docker", error);
     console.error("Stdout:\n", stdout);
     console.error("Stderr:\n", stderr);
     throw error;
   }
 }
+
+async function bundleLocally(
```
jmoseley, line 96: Did you try skipping the bundle step altogether, and doing a dynamic import? That would let us skip the ncc call altogether, at the cost of some parity with the regular build flow.

dereklegenzoff: The original implementation of the dev server, many commits ago, didn't do any bundling and just used the raw files so I'm pretty sure we can make that just fine if we like. I did this just to get closer to parity with the regular build flow. Although in the new world, the cost is much higher so not bundling would help the experience. 

Maybe we should make this a param so a user can choose to wait longer for a more "production like build" or just get the fast version?

packages/gensx/src/dev-server.ts
dereklegenzoff: fair, i'll add it to my list

packages/gensx/src/dev-server.ts
dereklegenzoff: Exactly - it's just for a better error message. I personally found the generic 404 confusing. I'll add in similar validation for the org.

packages/gensx/src/utils/bundler.ts
dereklegenzoff: Not really; just makes it more like the docker build step.  i'll remove it.

packages/gensx/src/utils/bundler.ts
dereklegenzoff: if we didn't have global there, wouldn't we end up modifying the user's package.json with this dependency? We could do `npx --package @vercel/ncc ncc build ./${relativeWorkflowPath} -o ${distDir} --target es2022` to run it without installing it but that adds a few seconds each time. 

Perhaps another argument for doing away with the bundling.

packages/gensx/src/utils/bundler.ts
dereklegenzoff: removed the extra code here

packages/gensx/src/dev-server.ts
dereklegenzoff: added validation for org as well. fortunately, it can all go in the same middleware function.

packages/gensx/src/dev-server.ts
dereklegenzoff: created an issue and assigned it to myself https://github.com/gensx-inc/gensx-console/issues/149

packages/gensx/src/utils/bundler.ts
dereklegenzoff: For posterity, jeremy and i chatted offline and decided to remove the bundle path completely; will just compile the files locally instead.

packages/gensx/src/utils/bundler.ts
dereklegenzoff: removed this code path altogether

jmoseley (approved):

PR #525, opened 2025-03-26
feat: Bundle with support for native dependencies.

jmoseley:
Use docker to bundle with support for native node addons.

TODO:
- [ ] Move docker image to its own repo and publish it


EvanBoyle (approved):
Overall looks good. I think we should avoid considering cloud build right now. Too many open questions and it is a lot of work to take on right now. 

I'm honestly fine with the docker dependency, I don't think it adds that much friction in practice. To make this as buttery smooth as possible we should probably run an upfront check ,`docker ps` or something, and return a well formatted message that indicates the user needs to install (or start) docker in order to build. 

I would love to understand a little better what it would take to remove the docker dependency. Is this really all about native dependencies? Having to use ncc and bundle the nodejs runtime into binary is a bit of a bummer from a coldstart perspective.

PR #530, opened 2025-03-31
fix: updating cli to match with new api shape

dereklegenzoff:
- removes `run` from the url path for running a workflow
- updates status to executionStatus
- removes stderror and stdout

### DIFF
--- a/packages/gensx/src/commands/run.ts
+++ b/packages/gensx/src/commands/run.ts
@@ -64,7 +64,7 @@ export async function runWorkflow(
 
       if (response.status >= 400) {
         throw new Error(
-          `Failed to start workflow: ${response.status} ${response.statusText}`,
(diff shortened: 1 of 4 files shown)


packages/gensx/src/commands/run.ts
```diff
@@ -145,19 +145,11 @@ export async function runWorkflow(
...
-        if (body.data.status === "failed") {
-          console.error("Workflow failed");
-          if (body.data.stderr) {
-            console.error("Stderr:", body.data.stderr);
-          }
-          if (body.data.stdout) {
-            console.error("Stdout:", body.data.stdout);
-          }
+        if (body.data.executionStatus === "failed") {
+          console.error("❌ Workflow failed");
```
jmoseley, line 152: When we have it, it might be nice to echo the command for checking the status.

jmoseley (approved):

PR #544, opened 2025-04-08
feat: make kind optional and fix getJson

dereklegenzoff:
- Make `kind` optional on BlobProvider. It defaults to "filesystem" if not provided unless process.env.GENSX_RUNTIME="cloud" in which case it defaults to "cloud".
- Also fixes a bug where getJson() wasn't properly parsing the json
- Adds tests for both

### DIFF
--- a/examples/chat-memory/src/workflows.tsx
+++ b/examples/chat-memory/src/workflows.tsx
@@ -84,11 +84,7 @@ const WorkflowComponent = gensx.Component<
   string
 >("Workflow", ({ userInput, threadId }) => (
   <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
-    <BlobProvider
-    //kind="filesystem"
-    //rootDir={path.join(process.cwd(), "chat-memory")}
-    //kind="cloud"
-    >
+    <BlobProvider>
       <ChatWithMemory userInput={userInput} threadId={threadId} />
     </BlobProvider>
   </OpenAIProvider>
--- a/examples/chat-memory/src/workflows.tsx
+++ b/examples/chat-memory/src/workflows.tsx
@@ -85,9 +85,9 @@ const WorkflowComponent = gensx.Component<
 >("Workflow", ({ userInput, threadId }) => (
   <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
     <BlobProvider
-      //kind="filesystem"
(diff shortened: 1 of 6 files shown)


examples/chat-memory/src/workflows.tsx
```diff
@@ -85,9 +85,9 @@ const WorkflowComponent = gensx.Component<
 >("Workflow", ({ userInput, threadId }) => (
   <OpenAIProvider apiKey={process.env.OPENAI_API_KEY}>
     <BlobProvider
-      //kind="filesystem"
-      //rootDir={path.join(process.cwd(), "chat-memory")}
-      kind="cloud"
+    //kind="filesystem"
+    //rootDir={path.join(process.cwd(), "chat-memory")}
+    //kind="cloud"
```
EvanBoyle, line 90: Let's just delete this.

packages/gensx-storage/src/blob/provider.tsx
```diff
@@ -23,16 +26,33 @@ import { BlobProviderProps } from "./types.js";
...
+    if (kind === "filesystem") {
+      // Ensure the storage directory exists
+      await mkdir(rootDir, { recursive: true });
+      const storage = new FileSystemBlobStorage(rootDir, props.defaultPrefix);
       return <BlobContext.Provider value={storage} />;
     } else {
       // Must be cloud based on our type definitions
-      const { defaultPrefix, organizationId } = props;
-      const storage = new RemoteBlobStorage(defaultPrefix, organizationId);
+      const organizationId =
```
EvanBoyle, line 50: Is this not also read from an environment variable? 

Executor should probably set `GENSX_ORG` and we should be able to read this from the ondisk config locally (or the same env var if the user sets it).

EvanBoyle: And is this actually an ID, or is it the organization name?

EvanBoyle (approved):

packages/gensx-storage/src/blob/provider.tsx
dereklegenzoff: Yeah i guess we should remove this, i'm not sure it's actually used. this was something that was in the original code that i didn't touch.

dereklegenzoff: removed; definitely a good improvement

PR #545, opened 2025-04-08
feat: Proper commonjs support

jmoseley:
## Proposed changes

Properly support commonjs.

* Rollup for bundling
* Use proper exports in package.json to support both ESM and CJS imports.
* Turn off `esModuleInterop` and no longer use `@/` shortcuts for typescript.
* Add tests for the compatibility examples so we can verify commonjs compatibility in CI.

Got some tips from https://evertpot.com/universal-commonjs-esm-typescript-packages/


EvanBoyle (commented):
Gave it a scan and overall makes sense, will give a full review once all of the tests are green. 
 
Does this impact dev server or `gensx deploy` at all? Are commonjs projects compatible with that surface area?

jmoseley:
> Does this impact dev server or gensx deploy at all?

It should work, but I need to some expansive testing on it once I get the tests sorted out.

jmoseley:
> Does this impact dev server or gensx deploy at all? Are commonjs projects compatible with that surface area?

CommonJS works with `gensx start` and deployments. I'm sure we will discover edge cases, but the major pieces are there.

EvanBoyle (approved):
Looks great, thanks for getting this taken care of and including docs!

Let's reach out to the users who were blocked on this once we've cut a new release.

PR #546, opened 2025-04-09
feat: Add DatabaseProvider and useDatabase hook

EvanBoyle:
Adds support for a `DatabaseProvider` and `useDatabase` hook.


examples/text-to-sql/src/data-ingestion.tsx
```diff
@@ -0,0 +1,80 @@
...
+        ('Marcus Bennett', 'Portland Pioneers', '1B', 550, 85, 25, 32, 98, 0.312, 0.385, 0.545, 0.930),
+        ('Ethan Carter', 'San Antonio Stallions', 'SS', 520, 92, 18, 24, 76, 0.298, 0.365, 0.512, 0.877),
+        ('Lucas Rivera', 'Charlotte Cougars', 'OF', 480, 78, 22, 28, 85, 0.285, 0.352, 0.498, 0.850),
+        ('Nathan Foster', 'Indianapolis Ironmen', '3B', 510, 75, 20, 26, 82, 0.301, 0.378, 0.525, 0.903),
+        ('Dylan Mitchell', 'Austin Archers', '2B', 490, 82, 15, 19, 65, 0.295, 0.362, 0.485, 0.847),
+        ('Jordan Hayes', 'Nashville Navigators', 'OF', 530, 88, 28, 35, 102, 0.308, 0.392, 0.558, 0.950),
+        ('Cameron Brooks', 'Orlando Owls', 'OF', 500, 76, 17, 22, 71, 0.292, 0.355, 0.495, 0.850),
+        ('Ryan Cooper', 'Sacramento Stars', 'C', 450, 65, 12, 15, 58, 0.278, 0.342, 0.465, 0.807),
+        ('Tyler Morgan', 'Memphis Monarchs', '1B', 540, 80, 23, 30, 95, 0.305, 0.375, 0.532, 0.907),
+        ('Brandon Reed', 'Las Vegas Legends', '3B', 510, 74, 19, 25, 78, 0.289, 0.358, 0.502, 0.860)
```
EvanBoyle, line 54: Small nit: this example will be much more compelling with more data. A hundred or a few hundred rows. Especially if we can find some public domain data (I assume this was generated with chatGPT). 

Though this is a fine start.

examples/text-to-sql/src/data-ingestion.tsx
```diff
@@ -0,0 +1,80 @@
...
+  "DatabaseInitializer",
+  async () => {
+    const DB_NAME = "baseball";
+
+    const storage = useDatabaseStorage();
+    let currentDatabases = await storage.listDatabases();
+
+    // If the database doesn't exist, populate it
+    // Otherwise, just return a message confirming it exists.
+    if (!currentDatabases.includes(DB_NAME)) {
```
EvanBoyle, line 19: Not the most bulletproof method as it relies on ordering. Another suggestion might be: 

- useDatabase
- run the `create table if not exists` (this is idempotent and safe to do every time)
- query the table to find the number of rows. 
- if zero rows, run the insert statement.

dereklegenzoff: sounds good, i will update to this

packages/gensx-storage/src/database/context.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Hook to access a database
+ * @param name The name of the database to access
+ * @returns A promise that resolves to a database object for the given name
+ * @throws Error if used outside of a DatabaseProvider
+ */
+export async function useDatabase(name: string): Promise<Database> {
+  const storage = useDatabaseStorage();
+
+  // Ensure the database exists before returning it
+  await storage.ensureDatabase(name);
```
jmoseley, line 37: `useDatabase` will be called a lot right? Basically anytime they use the database inside a component? It feels like this might add a lot of latency.

We can probably cache this, or something, so it just does it once, the first time the database is used inside the workflow.

dereklegenzoff: oh yeah that's a good point. Curious your thoughts on how we should approach this @EvanBoyle

packages/gensx-storage/src/database/types.ts
```diff
@@ -0,0 +1,284 @@
+// import { Readable } from "stream";
+
+import { InArgs } from "@libsql/client";
```
jmoseley, line 3:
```suggestion
import { InArgs } from "@libsql/client";
```

packages/gensx-storage/src/database/context.ts
EvanBoyle: Yes, it will be called whenever a component does 'useDatabase`. 

We could maintain a map of databases that have already been `used` in memory. But cache invalidation is always hard - remembering to delete from the cache on delete, etc. This is why I avoided storing clients in the API. We can add it now, or always do it later. I'm not strongly opinionated.

dereklegenzoff: went ahead and added support for this. We were already keeping a list of databases in memory so i just added a method to check if it's ensured and then i make sure we add it to the list when it's ensured and remove it when it's deleted.

jmoseley (approved):

PR #549, opened 2025-04-09
feat: Migrate to use the traces API.

jmoseley:
## Proposed changes

Use the `traces` API for creating traces. This uses the new model of `POST`/`PUT`.

### DIFF
--- a/packages/gensx-core/src/checkpoint.ts
+++ b/packages/gensx-core/src/checkpoint.ts
@@ -75,6 +75,9 @@ export class CheckpointManager implements CheckpointWriter {
   private consoleBaseUrl: string;
   private printUrl = false;
 
+  private traceId?: string;
+  private executionRunId?: string;
+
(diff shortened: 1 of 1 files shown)


packages/gensx-core/src/checkpoint.ts
```diff
@@ -75,6 +75,9 @@ export class CheckpointManager implements CheckpointWriter {
   private consoleBaseUrl: string;
   private printUrl = false;
 
+  private traceId?: string;
+  private executionRunId?: string;
```
EvanBoyle, line 79: Fine for now, but this is more global state that we'll have to rework when we move to durable.

EvanBoyle: (not that we won't have to do that anyways, since we're round tripping the entire JSON on every call from memory right now)

EvanBoyle (approved):

PR #554, opened 2025-04-10
chore: Fix repository that went missing from package.json

jmoseley:
## Proposed changes

The `repository` field got dropped from the `@gensx/openai` package.json somehow, so the publish is failing.


packages/gensx-openai/package.json
```diff
@@ -6,6 +6,11 @@
   "main": "./dist/cjs/index.cjs",
   "module": "./dist/esm/index.js",
   "types": "./dist/cjs/index.d.ts",
+  "repository": {
```
jmoseley, line 9: This got dropped during the commonjs rewrite

EvanBoyle (approved):

PR #557, opened 2025-04-11
fix: Fix bundling of storage package.

jmoseley:
## Proposed changes

Bundling is not compatible with @libsql/client, so work around it in the bundler by installing the package and specifying the correct native dependency.

https://github.com/tursodatabase/libsql-client-ts/issues/112

### DIFF
--- a/packages/gensx/src/utils/bundler.ts
+++ b/packages/gensx/src/utils/bundler.ts
@@ -42,7 +42,7 @@ export async function bundleWorkflow(
         "run",
         "--rm",
         "--platform",
-        "linux/amd64",
+        "linux/x86_64",
         "-v",
         `${packageJsonDir}:/app`,
         "-v",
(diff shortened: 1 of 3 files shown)


dereklegenzoff (approved):

packages/gensx/src/utils/bundler.ts
```diff
@@ -42,7 +42,7 @@ export async function bundleWorkflow(
         "run",
         "--rm",
         "--platform",
-        "linux/amd64",
+        "linux/x86_64",
```
EvanBoyle, line 45: ah, I suspected this might be incorrect?

EvanBoyle (approved):
Overall stinks to add an extra dep to every single build but I guess this is a fine workaround for now.

packages/gensx/src/utils/bundler.ts
jmoseley: I believe they are actually the same, but just to be sure.

jmoseley: image

PR #567, opened 2025-04-14
feat: add SearchProvider and useSearch hook

dereklegenzoff:
- adds support for `SearchProvider` and `useSearch` hooks
   - The "data plane" operations (upserting, querying, etc) match turbopuffer's SDK almost exactly with the exception of camelCasing param names
   - The "control plane" operations (list namespaces, delete namespace) are aligned with the blob/db methods
- adds a basic RAG example
- Also adds an OpenAIEmbedding component (courtesy of #513)

TODO:
- RAG example needs to be improved to have better data but can do that in a follow-up PR


packages/gensx-storage/src/search/context.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * @throws Error if used outside of a SearchProvider
+ */
+export async function useNamespace(name: string): Promise<Namespace> {
+  const search = useSearch();
+  // Only ensure the database if it hasn't been ensured before
+  if (!search.hasEnsuredNamespace(name)) {
+    await search.ensureNamespace(name);
+  }
+  return search.getNamespace(name);
+}
```
dereklegenzoff, line 40: We could consider changing these to `useSearchStorage` and `useSearch` to align better with the other providers. I'm on the fence.

EvanBoyle:
Let's do `useSearch(namespace)`

packages/gensx-storage/src/search/context.tsx
```diff
@@ -0,0 +1,40 @@
...
+ * Create the search context
+ */
+export const SearchContext = createContext<SearchStorage | null>(null);
+
+/**
+ * Hook to access the search instance
+ * @returns The search instance
+ * @throws Error if used outside of a SearchProvider
+ */
+export function useSearchStorage(): SearchStorage {
```
jmoseley, line 15: What is the use case for using this hook? Do we expect someone to always use `useSearch`?

packages/gensx-storage/src/search/remote.ts
```diff
@@ -0,0 +1,713 @@
...
+  }
+  throw new SearchNetworkError(`Error during ${operation}: ${String(err)}`);
+}
+
+/**
+ * Remote implementation of vector namespace
+ */
+export class SearchNamespace implements Namespace {
+  constructor(
+    public readonly id: string,
```
jmoseley, line 146:
```suggestion
    public readonly namespaceId: string,
```

examples/rag/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "node": ">=18.0.0"
+  },
+  "description": "Retrieval Augmented Generation with GenSX and the useSearch hook",
+  "scripts": {
+    "dev": "nodemon",
+    "start": "tsx ./src/index.tsx",
+    "build": "tsc",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "deploy": "npx gensx deploy -e OPENAI_API_KEY  -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e AWS_SESSION_TOKEN src/workflows.tsx"
```
jmoseley, line 16: What are the AWS creds for?

packages/gensx-storage/src/search/context.tsx
dereklegenzoff: This is the same setup used in both blob and database. `useSearchStorage` is more for control plane operations (list namespaces, create namespace) and `useSearch` is for data plane operations (upsert, query, etc). For blob it's `useBlob` and `useBlobStorage` and in databases it's `useDatabase` and `useDatabaseStorage`.

jmoseley: Gotcha. Do we expect users to use it? I'm worried about the documentation and confusion about which one to use. I guess same applies for the others too. But maybe its not a bit deal.

packages/gensx-storage/src/search/remote.ts
dereklegenzoff: updated

examples/rag/package.json
dereklegenzoff: fixed

packages/gensx-storage/src/search/context.tsx
dereklegenzoff: I think 90+% of the time you would use `useSearch` (and the same is true for blob and database). However, under the hood we're always using `useSearchStorage` under the hood on their behalf (go look at the implementation of useSearch). Someone will likely come up with a valid need to list search namespaces (or blobs or databases) in a workflow and `namespace.listNamespaces()` doesn't make much sense.

The concept probably makes the most sense for blob where the concepts of a `blob` vs `blobStorage` feel natural. It's not quite as natural for databases and search. But at the same time 

Maybe the reality is that control plane type operations should be provided through native ts/js support and we shouldn't export the extra hook? But there could be some limitations with that. Probably need to think more on this. I was gonna start in on the native ts/js support so could try to formulate a cohesive plan alongside that work.

jmoseley (approved):

packages/gensx-storage/tests/search/provider.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    // Database should be removed from cache
+    expect(search.hasEnsuredNamespace("test-db")).toBe(false);
+  });
+
+  test("defaultPrefix is used when provided", async () => {
+    let contextProvided = false;
+    let capturedStorage: SearchStorage | undefined;
+    const NestedComponent = gensx.Component<{}, null>("NestedComponent", () => {
+      const context = gensx.useContext(SearchContext);
```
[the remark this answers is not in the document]
dereklegenzoff: oh yeah good call. i'll update so we have a test that uses useSearch

PR #569, opened 2025-04-15
feat: adding native ts/js support for storage APIs

dereklegenzoff:
- Adding a `BlobClient`, `DatabaseClient`, and `SearchClient` that allow users to interact with storage without needing to use JSX.
- Updates the RAG and text-to-sql examples to use the new storage clients.
- Removed support for defaulPrefix in search and added a couple of control plane operations for blob for consistency.

### DIFF
--- a/packages/gensx-storage/src/blob/types.ts
+++ b/packages/gensx-storage/src/blob/types.ts
@@ -117,6 +117,13 @@ export interface BlobOptions {
 }
 
 /**
+ * Result of deleting a blob
+ */
+export interface DeleteBlobResult {
+  deleted: boolean;
+}
+
+/**
  * A response from the API
  */
 export interface BlobAPIResponse<T> {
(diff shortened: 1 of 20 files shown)


packages/gensx-storage/src/blob/types.ts
```diff
@@ -261,6 +268,16 @@ export interface BlobStorage {
    * List all blobs with the given prefix
    */
   listBlobs(prefix?: string): Promise<string[]>;
```
jmoseley, line 270: Just realized we forgot to paginate 😬

packages/gensx-storage/src/search/searchClient.ts
```diff
@@ -0,0 +1,67 @@
+// File: packages/gensx-storage/src/search/client.ts
+
```
jmoseley, line 2:
```suggestion
```

dereklegenzoff:
> We should add pagination to the list blobs endpoint, we will need it. But that can be another change.

Good point, I will add in pagination in a separate PR

packages/gensx-storage/src/search/searchClient.ts
dereklegenzoff: gotta love cursor's shennanigans

PR #571, opened 2025-04-15
fix: Use the right URL after deployments.

jmoseley:
## Proposed changes

Direct the user to the right place after deployments.


EvanBoyle (approved):
Should probably follow up and filter down to the right environment once that query param is availble.

PR #574, opened 2025-04-16
feat: updating CLI commands to use environments

dereklegenzoff:
- Adds `env` commands
   -  `gensx env create <name>`
   -  `gensx env ls`
   - `gensx env select <name>` - saves project config in `~/.config/gensx/projects/{projectName}.json`
   - `gensx env show` - reads from project config
   - `gensx env unselect` - removes environment from project config

Updates `start`, `deploy`, and `run` to support environments:
- `start`: tries to use value from `--env` then from cli then defaults to "default"
- `run` tries to use value from `--env` then from cli then throws an error
- `deploy` tries to use value from `--env` then from cli then asks for user input
   - if cli var is set, confirms with user that's the env they are about to use
   - if cli var is not set, has user pick from dropdown
   - if no env yet, has user type in an env name (defaults to "default")

Also updates environment variables to be specified as `-ev` and `--env-vars` instead of `-e` and `--env` to avoid confusion with environment name.

This also updates all of the dev server routes to just start with `/workflows`

Also adds test for any CLI command touched.


packages/gensx/src/index.ts
```diff
@@ -41,6 +41,7 @@ export async function runCLI() {
     .description("Start a local GenSX server")
     .argument("<file>", "File to serve")
     .option("-p, --project <name>", "Project name")
+    .option("--environment <name>", "Environment name")
```
dereklegenzoff, line 44: @jmoseley - FYI, i didn't include a short hand version of `--environment` because `-e` was already taken and i figure this will almost always be set via the gensx.yaml file.

Curious if you have any opinions here though

EvanBoyle (commented):
> Users can set the environment in the gensx.yaml file

This part does not sound right to me. You can have multiple environments but only a single gensx.yaml file. 

The way this should probably work is that environment state is managed for each project in a file in `~/.gensx/projects/project-name.json`

Then we can have an `environment` command `gensx environment create` `gensx environment ls` `gensx environment select`.

If you try to deploy without the --environment flag or env var set, we should prompt you to select an existing environment.

packages/gensx/src/index.ts
```diff
@@ -94,12 +101,51 @@ export async function runCLI() {
...
       .option(
         "-o, --output <file>",
         "Output file to write the workflow result to",
         undefined,
       )
       .action(runWorkflow);
+
+    // Environment management commands
+    const environmentCommand = program
+      .command("environment")
```
dereklegenzoff, line 114: Starting to lean towards making these `env` instead of environment. Feels like we'd then need to change how we specify env vars though (maybe -v and --envVar) 

Curious your thoughts @jmoseley

jmoseley: I'm ok with shorter. Definitely can't use `-v` (that is basically always "verbose"). What's wrong with `-e`?

packages/gensx/src/index.ts
jmoseley: oh right. that will be confusing.... `-e` and `--env` in the same command. Should have called them stacks 🤣 

but `--env` seems fine.

packages/gensx/src/dev-server.ts
```diff
@@ -270,46 +273,63 @@ export class GensxServer {
     this.app.use("*", cors());
 
     // Add organization and project validation middleware
-    this.app.use(`/org/:orgName/projects/:projectName/*`, async (c, next) => {
```
jmoseley, line 273: I'm having second thoughts about having these URLs all reflect the actual URL, as opposed to being just the suffix.

It feels weird to be specifying org, project, and environment (especially because its odd to "run a dev server for prod", but really its just a local server), just for the sake of constructing these URLs. An alternative would be to drop the prefix (eg `this.app.get("/workflows")`) and have it available on the project/environment page as a thing you copy and paste into your client config.

packages/gensx/src/index.ts
EvanBoyle: Yeah, I think `--env` would be an upgrade here. `-e` is used for exporting environment variables, no?

packages/gensx/src/dev-server.ts
EvanBoyle: My loose requirement here: 

If a user has a nextjs or client app that they are developing, they should be able to switch between developing against the dev server and developing against production workflows very easily. 

Unfortunately hard to understand what the e2e experience is without having an answer to https://github.com/gensx-inc/gensx/issues/579

I am happy with whatever we do here, as long as that loose requirement is satisfied. 

I do agree that having an org/project/stack in the URL for local development does feel odd. Users should not have to have an organization to use the dev server. Eventually, we will probably want to support building a bundle that contains the dev server so you can `node bundle.js` in your own container if you want to self-host the server. 

What do you think the right path is given all of that @jmoseley?

packages/gensx/src/index.ts
dereklegenzoff: Currently both `-e` and  `--env` are used for exporting environment variables. That's my dilemna here:
- if we change `gensx environment` to `gensx env` these commands have a different meaning for env than the deploy command does
- then for other commands (deploy, run, start), because we use both `-e` and  `--env` for environment variables in the deploy command, there's no good shorthand for `--environment`

EvanBoyle: Well easiest solution might be to change it to a different name. We could replace environment with `stack` and `--stack` and `-s`. But making that change would be painful at this point. 

@jmoseley any thoughts?

packages/gensx/src/dev-server.ts
dereklegenzoff: When i think about how i test things like express APIs locally, all i change is the base url and then things "just work". In my opinion, it's important that our local dev server should have an equivalent experience. That was my original motivation for making them long I suppose by that logic, we could consider everything before `workflows/{workflowName}` to be the base URL and use that rationale to shorten the routes.

jmoseley: > I suppose by that logic, we could consider everything before workflows/{workflowName} to be the base URL and use that rationale to shorten the routes.

That is what I'm thinking. There is probably a world where we start giving subdomains or short URLs as the base URL too.

I think that still fits the requirements for Evan.

dereklegenzoff: sounds good. I can get on board with this. i'll go make the updates.

dereklegenzoff: updates have been pushed

packages/gensx/src/index.ts
dereklegenzoff: updated environment to be -e and --env and environment variables to be -ev --env-var

packages/gensx/src/index.ts
dereklegenzoff: updated environment to be -e and --env and environment variables to be -ev --env-var

jmoseley:
> saves project config in ~/.gensx/projects/project-name.json

Why not use the existing config? Or at least a sibling location (`.config/gensx/<project-name>`)?

dereklegenzoff:
> > saves project config in ~/.gensx/projects/project-name.json
> 
> Why not use the existing config? Or at least a sibling location (`~/.config/gensx/<project-name>`)?

ahh sorry that comment is incorrect. Files get stored at `.config/gensx/projects/{projectName}.json` so they're in the same folder as the  main `.config/gensx/config` file

packages/gensx/src/commands/environment/unselect.ts
```diff
@@ -0,0 +1,52 @@
...
+import { validateAndSelectEnvironment } from "../../utils/env-config.js";
+import { readProjectConfig } from "../../utils/project-config.js";
+
+interface UnselectOptions {
+  project?: string;
+}
+
+export type { UnselectOptions };
+
+export async function handleUnselectEnvironment(options: UnselectOptions) {
```
EvanBoyle, line 14: Put this in the docs PR but I don't understand why we need this command.

packages/gensx/src/commands/environment/show.ts
```diff
@@ -0,0 +1,59 @@
...
+import { getSelectedEnvironment } from "../../utils/env-config.js";
+import { readProjectConfig } from "../../utils/project-config.js";
+
+interface ShowOptions {
+  project?: string;
+}
+
+export type { ShowOptions };
+
+export async function handleShowEnvironment(options: ShowOptions) {
```
EvanBoyle, line 14: IMO this command should just be `gensx env`

packages/gensx/src/commands/environment/unselect.ts
dereklegenzoff: The goal here was to strike the right balance between having the CLI be stateful (remember your selected env) while not having the state surprise users. Essentially balancing the preferences of you and @jmoseley. Without this command, a user could get locked into having the CLI be stateful. 

Also worth noting that pulumi offers the same command for their stack CLI: https://www.pulumi.com/docs/iac/cli/commands/pulumi_stack/

EvanBoyle: huh, TIL

EvanBoyle (approved):
I am good to go here. 

I'd still like to see `gensx env create` auto-select the stack by default. Can we do that before merging? We can always change the default behavior if we hear negative feedback in the future but I'm nearly certain this is the behavior users will expect (and anything else will be frustrating to use)

packages/gensx/src/commands/run.ts
```diff
@@ -45,10 +53,25 @@ export async function runWorkflow(
...
+      if (selectedEnvironment) {
+        environmentName = selectedEnvironment;
+        spinner.info(
+          `Using environment name from CLI: ${pc.cyan(environmentName)}`,
+        );
+      } else {
+        spinner.fail("No environment name provided");
+        throw new Error(
+          "No environment name found. Either specify --environment or run `gensx environment select <environment>`.",
+        );
```
jmoseley, line 67: Can this fallback to the selection/creation prompt flow?

dereklegenzoff: Yeah I can make that update. I was worried that it would be too annoying to see every time you ran something but maybe better safe than sorry.

packages/gensx/src/dev-server.ts
```diff
@@ -308,296 +274,278 @@ export class GensxServer {
...
-            createdAt: exec.createdAt,
-            finishedAt: exec.finishedAt,
-          }));
+    this.app.get(`/workflows/:workflowName/executions`, (c) => {
+      const workflowName = c.req.param("workflowName");
+      // Will throw NotFoundError if workflow doesn't exist
+      this.getWorkflowOrThrow(workflowName);
+
+      // Filter executions for this workflow
+      const executions = Array.from(this.executionsMap.values())
```
jmoseley, line 419: Should this hit our API if there is an API key?

packages/gensx/src/models/environment.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * List environments for a project
+ */
+export async function listEnvironments(
+  projectName: string,
+): Promise<ListEnvironmentsResponse["environments"]> {
+  const auth = await getAuth();
+  if (!auth) {
+    throw new Error("Not authenticated. Please run 'gensx login' first.");
```
jmoseley, line 24: Can we just offer to log them in here instead of an error? Not sure which UX is better.

packages/gensx/src/dev-server.ts
dereklegenzoff: I'm not sure i'm following. Why would the dev server hit our API?

jmoseley (approved):
I worry the tests are a bit heavy on the mocks, but otherwise this is generally good to go.

packages/gensx/src/dev-server.ts
jmoseley: Sorry I was confused. This is executions not traces.

packages/gensx/src/models/environment.ts
dereklegenzoff: I don't have a super strong opinion either way. I just followed the same convention that other commands used. 

Either way, if we make this update i'd prefer to save it for a separate PR.

packages/gensx/src/commands/run.ts
jmoseley: Well these are commands that only interact with our API, I'm not sure why you would run them without logging in...

I agree that we should not prompt/require it for running the dev server.

packages/gensx/src/models/environment.ts
jmoseley: Sounds good

packages/gensx/src/commands/run.ts
dereklegenzoff: updated this to fallback to the same selection/creation flow. refactored so they could use the same utility function.

packages/gensx/src/commands/environment/show.ts
dereklegenzoff: made the update. cursor seemed to agree with you.

PR #580, opened 2025-04-17
feat: adding pagination for list blobs

dereklegenzoff:
Adds pagination to the listBlobs methods. Uses cursor style pagination (`limit` and `cursor`).


packages/gensx-storage/src/blob/filesystem.ts
```diff
@@ -71,6 +73,35 @@ function getMetadataPath(filePath: string): string {
...
+function fromBase64UrlSafe(base64: string): string {
+  // Add back padding if needed
+  const pad = base64.length % 4;
+  const padded = pad ? base64 + "=".repeat(4 - pad) : base64;
+
+  return Buffer.from(
+    padded.replace(/-/g, "+").replace(/_/g, "/"),
+    "base64",
+  ).toString("utf-8");
+}
```
jmoseley, line 103: It feels funky to implement this ourselves, especially when we control both sides. Is there some other encoding scheme we can use for this?

packages/gensx-storage/src/blob/filesystem.ts
```diff
@@ -631,16 +662,16 @@ export class FileSystemBlobStorage implements BlobStorage {
...
-   * List all blobs with the given prefix
-   * @param prefix Optional prefix to filter blobs
-   * @returns An array of blob keys
+   * List all blobs with optional pagination
+   * @param options Options for listing blobs including prefix and pagination
+   * @returns A paginated list of blob keys
    * @throws BlobError if there's an error listing blobs
    */
-  async listBlobs(prefix?: string): Promise<string[]> {
+  async listBlobs(options?: ListBlobsOptions): Promise<ListBlobsResponse> {
```
jmoseley, line 670: Does this function fetch all blobs? Should we do it on demand instead as the user scrolls through the page or something? I expect users will have many many blobs.

packages/gensx-storage/src/blob/filesystem.ts
jmoseley: Actually, why is the client generating cursors itself? They should be opaque strings, owned by the server. I don't think there is a reason for the client to have any knowledge if the implementation of the cursors.

dereklegenzoff: this is the filesystem implementation of blobs so it never talks to the server--it is the server. I use the same scheme on the server side as well. 

In terms of not implementing it ourselves, it sounds like node might have `"base64url"` built in I can use to simplify this (or maybe it was a chatgpt hallucination, we'll find out)

packages/gensx-storage/src/blob/filesystem.ts
dereklegenzoff: by default this function will return 100 blob keys (unless overriden by the user). It supports the same pagination options as the server side (prefix, limit, nextCursor).

> Should we do it on demand instead as the user scrolls through the page or something? I expect users will have many many blobs.

I'm not quite sure what you're suggesting. Are you saying the implementation should be more efficient? Or we should return an iterator instead of the list?

packages/gensx-storage/src/blob/filesystem.ts
jmoseley: Ok I see that now. And yeah using the built in might be better. https://nodejs.org/docs/latest-v22.x/api/buffer.html#:~:text=Introduced-,base64url,-encoding.

dereklegenzoff: good call. I will make the update to that everywhere i'm encoding cursors

packages/gensx-storage/src/blob/filesystem.ts
jmoseley: Ah yeah I think I got confused about filesystem vs API.

jmoseley (approved):

PR #602, opened 2025-04-22
feat: GenSX Cloud MCP server

EvanBoyle:
Adds a new `@gensx/gensx-cloud-mcp` package with an stdio MCP server that enables connecting workflows deployed to a particular environment to an MCP client like Claude desktop, Cursor, etc. 

- Basic implementation only allows connecting a single environment, though users should be able to add multiple entries to their MCP config to connect other projects as well. 
- Auth is read from gensx config file by default, with the option to provide an environment variable in the config directly
- Added a short docs page


packages/gensx-cloud-mcp/CHANGELOG.md
```diff
@@ -0,0 +1,158 @@
+# Changelog
```
dereklegenzoff, line 1: this needs to be removed

EvanBoyle: Removed the content but kept the file. I assume this is right?

website/docs/src/content/cloud/mcp-server.mdx
```diff
@@ -0,0 +1,55 @@
...
+      ],
+      "env": {
+        "GENSX_API_KEY": "my_api_key"
+      }
+    }
+  }
+}
+```
+
+The GenSX build process automatically extracts input and output schemas from your typescript types, so no additional configuration or manual `zod` schema is required to consume your workflows from an MCP server.
```
dereklegenzoff, line 55: This doc feels incomplete. How do i run the MCP server once i update this config? What should i expect to see when i run it? How do i hook it up to claude code or another tool?

packages/gensx-cloud-mcp/src/index.tsx
```diff
@@ -0,0 +1,387 @@
...
+    console.error("Using API key from environment variable");
+    return envApiKey;
+  }
+
+  // Fall back to config file
+  const config = readConfig();
+  if (config.api?.token) {
+    console.error("Using API key from config file");
+    return config.api.token;
+  }
```
dereklegenzoff, line 35: why is this console.error rather than console.info?

packages/gensx-cloud-mcp/src/index.tsx
```diff
@@ -0,0 +1,387 @@
...
+          if (!workflow.inputSchema) {
+            console.error(
+              `Workflow ${workflow.name} has no input schema, fetching details...`,
+            );
+            const detailedWorkflow = await client.getWorkflow(workflow.name);
+            workflow.inputSchema = detailedWorkflow.inputSchema;
+            workflow.outputSchema = detailedWorkflow.outputSchema;
+          }
+
+          // Create a simple validator function based on required fields in the schema
```
dereklegenzoff, line 238: In the API, we use the `ajv` library for this. Probably worth just re-using the same approach for consistency

```
import { Ajv, ErrorObject } from "ajv";

// Function to validate input against JSON schema
export function validateWorkflowInput(
  input: unknown,
  schema: unknown,
): string | null {
  if (!schema) return null; // No schema means no validation errors

  try {
    const parsedSchema =
      typeof schema === "string" ? JSON.parse(schema) : schema;

    const validate = ajv.compile(parsedSchema);
    const valid = validate(input);

    if (!valid) {
      const errors = validate.errors ?? [];
      const formattedErrors = errors.map((err: ErrorObject) => {
        return `${err.instancePath} ${err.message}`.trim();
      });

      return formattedErrors.join("; ");
    }
  } catch (err) {
    console.error("Schema validation error:", err);
  }

  return null;
}
```

packages/gensx-cloud-mcp/src/index.tsx
```diff
@@ -0,0 +1,387 @@
...
+
+    const server = await setupServer();
+    const transport = new StdioServerTransport();
+    await server.connect(transport);
+
+    console.error(
+      `GenSX Cloud MCP Server running on stdio for ${org}/${projectName}/${environmentName}`,
+    );
+  } catch (error) {
+    console.error("Fatal error running server:", error);
```
dereklegenzoff, line 379: console.info?

packages/gensx-cloud-mcp/CHANGELOG.md
dereklegenzoff: i'm not entirely sure. you can probably just fully delete it but i don't know if it matters either way

website/home/public/llms.txt
```diff
@@ -16,34 +16,40 @@ GenSX is a simple typescript framework for building complex LLM applications. It
 - [Model Context Protocol](https://gensx.com/docs/examples/mcp)
```
dereklegenzoff, line 16: one of these day's ill get around to removing these from the gitignore :)

dereklegenzoff (approved):
Docs need a little love and you seem to be using console.error a bunch of places you shouldn't be but otherwise looks good

packages/gensx-cloud-mcp/src/index.tsx
EvanBoyle: Claude splits up logs. I just put everything into the err log to make things easier to debug.

packages/gensx-cloud-mcp/src/index.tsx
EvanBoyle: Will merge as is an revisit this when I make subsequent updates to the new API shape.

website/docs/src/content/cloud/mcp-server.mdx
EvanBoyle: Understood, will spruce these up more when we launch it in a couple of weeks. 

Configuring an MCP server is the same across any tool. There is a JSON file with MCP server config and you just update it to include the JSON config above. It is a non-goal for us to document every tool like claude/cursor, etc and their specific settings.

packages/gensx-cloud-mcp/CHANGELOG.md
jmoseley: You can fully delete it, but leaving it here is fine too.

packages/gensx-cloud-mcp/src/index.tsx
```diff
@@ -0,0 +1,387 @@
...
+    console.error(
+      `Found ${workflows.length} workflows in environment ${environmentName}`,
+    );
+
+    // Convert workflows to tools
+    const workflowTools = await Promise.all(
+      workflows.map(async (workflow) => {
+        try {
+          // For workflows without a schema, fetch detailed information
+          if (!workflow.inputSchema) {
```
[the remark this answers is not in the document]
EvanBoyle: No this is just defensive coding. I can delete and throw an error instead if you think we should.

website/docs/src/content/cloud/mcp-server.mdx
dereklegenzoff: I'm not suggesting we document every tool (although showing one would be nice).

I'm more saying that looking at this doc i don't understand how to use this.  do i need to run a command to run the gensx-cloud-mcp server? Or just because this package exists on npm now i can just update my mcp config? maybe i'm just missing something

dereklegenzoff: or is there a server side element that hasn't been built yet?

EvanBoyle: This config file tells the tool in question to run `npx @gensx/gensx-cloud-mcp` which will start the server on behalf of the user. That's it. The package gets acquired from NPM automatically.

dereklegenzoff: ahhh i see. that makes so much more sense

PR #604, opened 2025-04-22
docs: updating CLI docs and other references to CLI commands + dev server

dereklegenzoff:
- Adds CLI reference for `env` commands
- Updates CLI reference for other commands
- Updates dev server routes and CLI commands in conceptual docs


EvanBoyle (commented):
@dereklegenzoff you'll need to create a `_meta.ts` file for the cli-reference section. The new env commands folder isn't showing up in the nav.

dereklegenzoff:
> @dereklegenzoff you'll need to create a `_meta.ts` file for the cli-reference section. The new env commands folder isn't showing up in the nav.

ahh good catch

website/docs/src/content/cli-reference/env/show.mdx
```diff
@@ -0,0 +1,38 @@
+---
+title: gensx env show
+description: Show current GenSX environment details
+---
+
+# gensx env show
+
+The `gensx env show` command displays the name of the currently selected environment.
+
```
EvanBoyle, line 9: Should this just be `gensx env` to show the currently selected env? This is fine to have as well I guess.

website/docs/src/content/cli-reference/env/unselect.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: gensx env unselect
+description: Unselect the current GenSX environment
+---
+
+# gensx env unselect
```
EvanBoyle, line 6: Why is this command necessary? Feels like we should delete it from the CLI.

website/docs/src/content/cli-reference/index.mdx
```diff
@@ -17,29 +17,65 @@ npm install -g gensx
...
 
-# Start local development server
+# Start local development
 gensx start src/workflows.tsx
+```
+
+### Managing Environments
+```bash
+# Create and switch to a development environment
+gensx env create dev
```
EvanBoyle, line 65: Should `create` automatically select the env? Feels liek the default we would want.

EvanBoyle (approved):
A few small comments but overall looks good. Let's make sure to add a _meta file for the CLI reference and make sure all command pages are rendering before we merge.

website/docs/src/content/cli-reference/index.mdx
dereklegenzoff: The goal here was to strike the right middle ground between having the CLI be stateful and giving users explicit control. If we had this select the env, we'd be automatically opting people into the statefulness. 

One option would be to add a `--select` option as a middle ground.

EvanBoyle: Both explicit select and --select flag feels really painful here. The 99% case when you create an environment is that you want to immediately use it. 

The only time someone could potentially ever get surprised is when you `deploy` or `run` and we prompt you to confirm anyways right?

dereklegenzoff: When we chatted about this in slack last week, one of the principles we agreed on in the thread was: "We never 'select' an environment on your behalf, you have to do it explicitly". I'll make the update if you want but it would go against what we had previously agreed on.

EvanBoyle: That was not my takeaway. no surprises and no mutating anything production related (`gensx deploy` `gensx run`) based on implcit env config, just prompt the user to confirm.

dereklegenzoff: I copied that verbatim from the thread :)

Anyways, i will go make the update.

EvanBoyle: Apologies for attempting to rewrite history 😆 

Thanks for making the change

PR #610, opened 2025-04-23
fix: Ensure that we wait for pending checkpoints after streaming workflow output finishes

jmoseley:
## Proposed changes

Ensure that we call `waitForPendingUpdates` _after_ the entire output stream has been consumed. This ensure that we actually wait for the final checkpoints, which may not be created until the stream fully flushes.


EvanBoyle (approved):
Not sure this will solve the bugs I've been seeing. 

I guess I am technically using a StreamComponent in non-streaming (which might still use a stream under the hood): 

```
import * as gensx from "@gensx/core";
import { ChatCompletion, OpenAIProvider } from "@gensx/openai";

const WebResearchComponent = gensx.Component<{ query: string }, string>(
  "WebResearchComponent",
  ({ query }) => {
    return (
      <OpenAIProvider
        apiKey={process.env.PERPLEXITY_API_KEY}
        baseURL={"https://api.perplexity.ai"}
      >
        <ChatCompletion
          messages={[
            {
              role: "system",
              content:
                "You are a helpful assistant that can answer questions about the web. Include URL citations in the text of the response",
            },
            { role: "user", content: query },
          ]}
          model="sonar-pro"
        />
      </OpenAIProvider>
    );
  }
);

const WebResearchWorkflow = gensx.Workflow(
  "WebResearchWorkflow",
  WebResearchComponent
);

export { WebResearchWorkflow };
```

jmoseley:
This has turned into a larger issue. When streaming, we do not properly await the conclusion of child nodes either. Going to put a pin in this for a bit.

EvanBoyle:
@jmoseley can you open a bug with your finding so we don't lose context?

jmoseley:
#628

PR #618, opened 2025-04-23
feat: add bin to mcp server package.json

EvanBoyle:
MCP server needs a `bin` entrypoint similar to the CLI.


packages/gensx-cloud-mcp/package.json
```diff
@@ -6,6 +6,9 @@
   "main": "./dist/cjs/index.cjs",
   "module": "./dist/esm/index.js",
   "types": "./dist/cjs/index.d.ts",
+  "bin": {
+    "gensx-cloud-mcp": "./dist/esm/index.js"
```
jmoseley, line 10: Since this is a CLI package, we don't need to support dual ESM/CJS builds. See https://github.com/gensx-inc/gensx/blob/main/packages/gensx/package.json for an example of a package like that.

jmoseley (approved):

jmoseley:
Did you check that this will run with `npx` as expected? Would it make more sense to move this package into `gensx`? `npx gensx mcp ...`

EvanBoyle:
@jmoseley it could make sense to move this into the CLI at some point, but there is no reason for a user to ever run this command on their own. Given that i think it probably makes more sense to keep it separate.

packages/gensx-cloud-mcp/package.json
EvanBoyle: Ok, removed all of the rollup and cjs stuff to match the CLI build.

PR #620, opened 2025-04-23
fix: adding user agents

dereklegenzoff:
Adds user agents for gensx-storage, gensx-core, and gensx-cloud-mcp


EvanBoyle (approved):
Do we have a user agent for the CLI as well?

jmoseley (approved):

dereklegenzoff:
> Do we have a user agent for the CLI as well?

Yeah, we already have user agents in the CLI

Issue #486, opened 2025-03-18
bug: `gensx new` creates yaml file with undefined description


EvanBoyle:
running `npx gensx new .` creates a yaml file with an `undefined` description. We should just omit this, use a default, or prompt the user to supply a value. 

![Image](https://github.com/user-attachments/assets/096c738b-0fcb-4565-b23e-91e3dd44bfa5)

PR #641, opened 2025-04-29
feat: wiring up support for description in gensx.yaml

### LINKED ISSUE #582: Add description to `gensx.yaml` project file
We should add support for a description field in the gensx.yaml file, and render this description in the console.

dereklegenzoff:
- Prompts user for a description in the `new` command if not provided (can just hit enter to skip)
- Removes `description` from gensx.yaml in starter project if it's not defined (avoids the `description=undefined`
- Adds description to the createProject model
- Includes description in the deploy command

Fixes #582 and #486


EvanBoyle:
> Removes description from gensx.yaml in starter project if it's not defined

Might be worth adding a default description (I believe other tools to this) to aid with discoverability of the feature (users will know they can change it).

packages/gensx/src/commands/deploy.ts
```diff
@@ -79,6 +79,10 @@ export async function deploy(file: string, options: DeployOptions) {
 
     form.append("schemas", JSON.stringify(schemas));
 
+    if (projectConfig?.description) {
+      form.append("description", projectConfig.description);
```
jmoseley, line 83: Do we need to update the API?

dereklegenzoff: yes, there is a sister PR for the API here: https://github.com/gensx-inc/gensx-console/pull/254/files

dereklegenzoff:
> > Removes description from gensx.yaml in starter project if it's not defined
> 
> Might be worth adding a default description (I believe other tools to this) to aid with discoverability of the feature (users will know they can change it).

Probably a good call. i'll make the update

jmoseley (approved):

PR #647, opened 2025-04-30
feat: scoping storage to project/env

dereklegenzoff:


packages/gensx-core/src/utils/project-config.ts
```diff
@@ -0,0 +1,51 @@
...
+});
+
+export type ProjectConfig = z.infer<typeof ProjectConfigSchema>;
+
+/**
+ * Get the path to the gensx.yaml file
+ */
+export function getProjectConfigPath(dir: string): string {
+  return path.join(dir, "gensx.yaml");
+}
```
EvanBoyle, line 20: I wonder if the logic for finding a project file is robust enough. I think this should be fine as all of our examples have the project file at the same level as the package.json where a user will be running commands (the directory we'll detect as entrypoint). But wouldn't be surprised if we discover some edge cases that require recursively searching parent directories for this file. Probably fine for now.

EvanBoyle (approved):
Overall everything looks good to me. 

Surprised to not see `use` functions updated to take an options bag to optionally specify scope inline. Configuring a provider is the only way to change scope. That makes using resources from other envs more painful.

EvanBoyle:
Also remembering that we need to go in and update the docs for both storage and projects

packages/gensx-core/src/utils/project-config.ts
jmoseley: Agree, this will definitely bite someone.

packages/gensx-core/src/utils/project-config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Read the gensx.yaml file and return the parsed config
+ */
+export function readProjectConfig(dir: string): ProjectConfig | null {
```
jmoseley, line 25: I feel like we added support for JSON or yaml. Maybe that was a fever dream.

packages/gensx-storage/src/blob/types.ts
```diff
@@ -327,6 +327,16 @@ export interface BaseBlobProviderProps {
    * Default prefix for all blob keys
    */
   defaultPrefix?: string;
+
+  /**
```
jmoseley, line 331: We should improve the jsdocs to talk about the defaults.

packages/gensx-storage/src/database/types.ts
```diff
@@ -253,6 +253,16 @@ export interface BaseDatabaseProviderProps {
    * Storage kind
    */
   kind?: DatabaseStorageKind;
+
+  /**
+   * Project name
```
jmoseley, line 258: Ditto, mention the defaults

dereklegenzoff:
> Also remembering that we need to go in and update the docs for both storage and projects

I'll go make the updates as a follow up. From a code perspective, nothing really changes which is nice. 

> Surprised to not see use functions updated to take an options bag to optionally specify scope inline. Configuring a provider is the only way to change scope. That makes using resources from other envs more painful.

In the doc i shared yesterday on some implementation details, others weren't a fan of supporting this in `use` calls. I left it out for now for expediency and we can always add it in later if we need.

packages/gensx-core/src/utils/project-config.ts
dereklegenzoff: I added an issue to look into this: https://github.com/gensx-inc/gensx/issues/648

packages/gensx-core/src/utils/project-config.ts
dereklegenzoff: this is identical to the code from the CLI (albeit sync instead of async). Looks like only yaml is supported.

packages/gensx-storage/src/database/types.ts
dereklegenzoff: good call, updated

packages/gensx-core/src/utils/project-config.ts
jmoseley: Ok. Hallucination then lol. Should this use the same code as the CLI?

dereklegenzoff: is there a recommended way to share between `gensx` and `gensx-core`? the other config stuff had just been duplicated between those two folders already but i agree it'd be nicer to only have the code in one place

PR #670, opened 2025-05-06
feat: upgrading env cli commands to use ink

dereklegenzoff:
First set of updates to migrate commands to ink:
- updates all `env` commands
- updates tests to use "ink-testing-library" and do less mocking
- also tried to land on some best practices like separating logic out into a hook from the main component

Short video:
https://screen.studio/share/vN0TicO3


packages/gensx/tests/commands/environment/create.test.ts
```diff
@@ -1,30 +1,40 @@
...
+  afterEach,
+  beforeAll,
+  beforeEach,
+  expect,
+  it,
+  suite,
+  vi,
+} from "vitest";
+
+// Import with the correct file extension
```
jmoseley, line 16: Hello cursor

dereklegenzoff: lol yeah probably need to go and clean up some of the comments from cursor

packages/gensx/tests/commands/environment/list.test.ts
```diff
@@ -24,27 +38,40 @@ vi.mock("../../../src/models/projects.js", () => ({
...
+  // Set working directory to our test project
+  process.cwd = vi.fn().mockReturnValue(path.join(tempDir, "project"));
 });
 
-afterEach(() => {
-  console.info = originalConsoleInfo;
+afterEach(async () => {
   vi.resetAllMocks();
+  await cleanupProjectFiles(tempDir);
 });
```
jmoseley, line 61: I think all this test setup could be moved into a setup.ts file. https://vitest.dev/config/#setupfiles

dereklegenzoff: fair, i can do that. i refactored a decent bit of the code into a helpers fule but agreed the shared setup would be best.

dereklegenzoff: actually i'll do this refactoring with my next PR that upgrades everything else to ink; otherwise i'll end up refactoring twice.

PR #673, opened 2025-05-08
docs: fixing build of website

dereklegenzoff:
Adding react/types to pnpm-workspace to ensure a consistent version of react types. fixes build that was broken by introducing ink (react 18) in the CLI


packages/gensx/package.json
```diff
@@ -44,7 +44,7 @@
     "@types/chokidar": "^2.1.7",
     "@types/ini": "^4.1.1",
     "@types/node": "catalog:packages",
-    "@types/react": "^18.2.0",
+    "@types/react": "catalog:",
```
jmoseley, line 47: we should probably just use whatever makes sense for the project instead of sharing this across all projects using the catalog. I assume we will update react separately per project.

website/docs/package.json
```diff
@@ -28,7 +28,7 @@
   "devDependencies": {
     "@tailwindcss/typography": "^0.5.16",
     "@types/node": "^20.17.30",
-    "@types/react": "^19",
+    "@types/react": "catalog:",
```
jmoseley, line 31: ditto, can we not just pin this to the right thing?

jmoseley (approved):
I'm hesitant to share the react version across all of our frontends, unless we think there is a good reason to have them always be on the same version. But I will leave it up to you.

packages/gensx/package.json
dereklegenzoff: sounds good, i will just update them individually

PR #677, opened 2025-05-12
feat: updating remaining CLI commands to use ink

dereklegenzoff:
Updates the remaining CLI commands to use ink. Also improved the test setup process and reduced tests dependence on mocks (a bit).

Also makes some other minor improvement like displaying the error of a failed `run`.

## Run 

https://github.com/user-attachments/assets/7a824a80-a7ec-4f11-94fd-7eb92512f927

## Deploy

https://github.com/user-attachments/assets/e131a688-f21e-4bc5-b42b-13fc690e4d4b

## New

https://github.com/user-attachments/assets/53209db7-9dab-4750-b506-b511a185471a

## Build

https://github.com/user-attachments/assets/1ec95c76-965e-438b-b693-3a9ec4cfcfa3

## Login

https://github.com/user-attachments/assets/bea39d40-fcdd-4a5e-bac8-9cf54bf42ea4

## Dev server

https://github.com/user-attachments/assets/35024588-5dd4-428d-9dfb-ff1727c344aa


packages/create-gensx/tests/package.test.ts
```diff
@@ -99,6 +100,7 @@ it("package.json is correctly configured for npm create", async () => {
   const testProjectDir = path.join(tempDir, "test-project");
 
   try {
+    // Set NODE_ENV to production to avoid React development mode warnings
```
jmoseley, line 103:
```suggestion
```

packages/create-gensx/tsconfig.json
```diff
@@ -1,10 +1,29 @@
...
+    "inlineSources": true,
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noFallthroughCasesInSwitch": true,
+    "jsx": "react-jsx",
+    "jsxImportSource": "react"
   },
-  "include": ["src/**/*", "vitest.config.ts", "tests/**/*"],
+  "include": ["src/**/*", "tests/**/*", "vitest.config.ts"],
```
jmoseley, line 27: How much of this is necessary?

jmoseley (approved):
Seems good! 🖊️ 

Make sure we test `npx create gensx` and `brew install gensx` when we are done.

packages/create-gensx/tests/package.test.ts
dereklegenzoff: ahh cursor :)

packages/create-gensx/tsconfig.json
dereklegenzoff: Good question. I just pasted this to match exactly what was in the `tsconfig.base.json` that we can no longer reference here (because we need jsx from react).

I could go through and clean it up but perhaps it's best just to be consistent with other packages?

jmoseley: Yeah consistency is fine. Can't you still import the base config and update the JSX options?

dereklegenzoff: oh yeah that works great. updated both tsconfig files

PR #682, opened 2025-05-14
fix: making user agent logic permissive

dereklegenzoff:
adding try/catch so that the version will just be unset if package.json can't be found. this will fix errors with using gensx via homebrew.

### DIFF
--- a/packages/gensx/src/utils/user-agent.ts
+++ b/packages/gensx/src/utils/user-agent.ts
@@ -4,9 +4,15 @@ import { fileURLToPath } from "node:url";
 
 const dirname = path.dirname(fileURLToPath(import.meta.url));
 
-const localPackageJson = JSON.parse(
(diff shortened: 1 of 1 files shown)


jmoseley:
Will the user agent just be broken for brew installations?

dereklegenzoff:
> Will the user agent just be broken for brew installations?

yeah we just won't get the version (similar to how we also don't get user agents from deployed workflows)

jmoseley (approved):
I guess this is fine, but a bit disappointing. Means we might get a lot of `unknown` for versions and make it hard to track whats in use out there.

dereklegenzoff:
> I guess this is fine, but a bit disappointing. Means we might get a lot of unknown for versions and make it hard to track whats in use out there.

Yeah i agree we should find a better approach and i think we already have an issue open tracking this. goal of this PR is just to unblock homebrew installations.

PR #692, opened 2025-05-20
feat: removing providers and introducing useBlob/search/db helpers

dereklegenzoff:
This PR removes providers/contexts in @gensx/storage in favor of hooks: useBlob, useSearch, useDatabase.

- Also updates tests and docs
- Renames `BlobProviderProps` to `BlobStorageOptions` and equivalent for other storage modalities


jmoseley (approved):
We should probably clean up those tests that check for function types but 🤷

PR #695, opened 2025-05-20
feat: Decorators instead of JSX.

jmoseley:
## Proposed changes

The first pass at decorator support. Still need to fix linting.


examples/hacker-news-analyzer/tsconfig.json
```diff
@@ -10,11 +11,14 @@
...
+    "outDir": "./dist",
+    "plugins": [
+      {
+        "name": "ts-function-decorator-ls"
+      }
+    ]
   },
-  "include": ["./src/**/*.ts", "./src/**/*.tsx"],
-  "exclude": ["node_modules"]
+  "include": ["./*.ts", "./*.tsx", "./**/*.ts", "./**/*.tsx"],
```
dereklegenzoff, line 22: I assume we no longer need tsx?
```suggestion
  "include": ["./*.ts", "./**/*.ts"],
```

examples/hacker-news-analyzer/package.json
```diff
@@ -7,22 +7,23 @@
...
-    "dev": "tsx ./src/index.tsx",
-    "start": "npx gensx start ./src/workflows.tsx",
-    "build": "tsc",
+    "dev": "nodemon",
+    "start": "pnpm build && node dist/index.js",
+    "build": "tsc -p tsconfig.build.json",
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
-    "deploy": "npx gensx deploy -ev OPENAI_API_KEY ./src/workflows.tsx"
+    "prepare": "ts-patch install -s"
```
dereklegenzoff, line 15: Probably worth discussing what the ideal set of commands to have here is. I had gone through and updated this example so that `start` ran the dev server and `deploy` would do the deployment

dereklegenzoff: is running `pnpm prepare` a prereq now?

jmoseley: Yes for projects that use decorators.

jmoseley: We can definitely replace nodemon with the dev server.

packages/gensx-core/src/context.ts
```diff
@@ -1,5 +1,4 @@
-import { resolveDeep } from "./resolve.js";
-import { ComponentProps, Context, GsxComponent } from "./types.js";
+import { ComponentProps, Context } from "./types.js";
```
dereklegenzoff, line 1: is context an internal feature now or still a customer facing feature?

examples/hacker-news-analyzer/package.json
dereklegenzoff: I had done it the other way becasue it felt weird for `pnpm dev` to run `gensx start`. Could always change the CLI command name though

dereklegenzoff (approved):
🚀

packages/gensx-core/src/context.ts
jmoseley: Yes internal. Basically used for tracking state of parent/child nodes.

examples/hacker-news-analyzer/package.json
jmoseley: We could make it `gensx serve` or something.

packages/gensx-core/src/index.ts
```diff
@@ -1,21 +1,5 @@
...
-  ComponentProps,
-  GsxStreamComponent,
-  GsxComponent,
-  GSXToolProps,
-  GSXToolAnySchema,
-} from "./types.js";
-export type { GsxArray } from "./array.js";
-export { readConfig } from "./utils/config.js";
-export { getSelectedEnvironment } from "./utils/env-config.js";
-export { readProjectConfig } from "./utils/project-config.js";
```
dereklegenzoff, line 21: oh just realized we are still gonna need these. they're used by the storage package

PR #697, opened 2025-05-21
fix: Remove linting from examples for now.

jmoseley:
## Proposed changes

Remove and disable eslint for examples. I was not able to overcome the eslint error inside the timebox.


dereklegenzoff (approved):

dereklegenzoff (commented):
should we also remove the eslint dependency in all the package.json's?

jmoseley:
> should we also remove the eslint dependency in all the package.json's?

If its there, I thought it was only at the top level, but yes remove it if you come across it in the examples. I want to keep eslint for the packages, it does provide value there.

PR #701, opened 2025-05-22
feat: Add wrap helpers

jmoseley:
## Proposed changes

* `wrap` helper that can wrap an arbitrary instance
* `Wrap` decorator that can decorate a class declaration
* `wrapOpenAI` helper for wrapping an openAI instance (with deep wrapping of tool calls for `runTools`)
* exported `OpenAI` class that is a drop in replacement so you can just swap `import { OpenAI } from 'openai';` with `import { OpenAI } from '@gensx/openai';`
* Update docs


.nvmrc
```diff
@@ -1 +1 @@
-22
+20
```
dereklegenzoff, line 1: was this intentional?

jmoseley: yes, gets rid of the punycode warning 🤣

examples/blog-writer/src/workflows.ts
```diff
@@ -0,0 +1,181 @@
...
+      //   name: "research",
+      //   description: "Research a topic",
+      //   parameters: z.object({
+      //     topic: z.string(),
+      //   }),
+      //   function: (args: { topic: string }) => {
+      //     console.log("🔍 Research tool:", args.topic);
+      //     return `You researched the topic ${args.topic} and found the following: ${args.topic}`;
+      //   },
+      // }),
```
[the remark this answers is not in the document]
jmoseley: there are two ways to specify functions, but the types don't allow you to mix and match, and I wanted to test both.

dereklegenzoff (approved):
looks great!!

.nvmrc
dereklegenzoff: ooh that is exciting. i really hate that warning.

PR #704, opened 2025-05-22
docs: updating openai and anthropic examples

dereklegenzoff:
neither are quite at full parity as the old versions, especially anthropic but we can improve them in the future.


examples/openai-examples/package.json
```diff
@@ -8,20 +8,23 @@
...
-    "build": "tsc"
+    "dev": "pnpm build && node dist/index.js",
+    "start": "npx gensx start ./src/workflows.ts",
+    "build": "tsc -p tsconfig.build.json",
+    "deploy": "npx gensx deploy -e OPENAI_API_KEY ./src/workflows.ts",
+    "prepare": "ts-patch install -s"
   },
   "dependencies": {
     "@gensx/openai": "workspace:*",
     "@gensx/core": "workspace:*",
```
jmoseley, line 19: Should we change these to alpha?

PR #705, opened 2025-05-23
feat: updating examples and supporting comp/wf with no inputs

dereklegenzoff:
- updates `@gensx/core` to support workflows/components with no inputs
- also fixes wrap to support functions/methods with no inputs (previously through an error)
- updates deep research and all-models-metadata examples
- updates text-to-sql to no longer need empty inputs


examples/deep-research/src/firecrawlProvider.ts
jmoseley, line 1: nit: new filename?

packages/gensx-core/src/wrap.ts
```diff
@@ -52,9 +61,15 @@ export function wrap<T extends object>(sdk: T, opts: WrapOptions = {}): T {
             [...path, String(propKey)].join(".");
 
           // Bind the original `this` so SDK internals keep working
-          const boundFn = value.bind(origTarget) as (input: object) => unknown;
-          const componentOpts = opts.getComponentOpts?.(path, boundFn);
-          return wrapFunction(boundFn, {
+          const boundFn = (value as (...args: unknown[]) => unknown).bind(origTarget);
+
+          // Handle functions with no parameters vs functions with optional parameters
+          const wrappedFn = boundFn.length === 0
```
jmoseley, line 67: 🤯 this is new to me, I think I have some other places this will be useful.

packages/gensx-core/src/wrap.ts
```diff
@@ -112,13 +127,20 @@ export function wrap<T extends object>(sdk: T, opts: WrapOptions = {}): T {
...
     componentOpts?.name ?? (fn.name || "AnonymousComponent");
-  return createComponent(fn, { ...componentOpts, name: componentName });
+
+  // Create a wrapper that properly handles Promise unwrapping
+  const wrappedFn = async (input?: TInput): Promise<Awaited<TOutput>> => {
+    const result = await fn(input!);
+    return result;
+  };
+
+  return createComponent(wrappedFn, { ...componentOpts, name: componentName }) as (props?: TInput, componentOpts?: ComponentOpts) => Promise<Awaited<TOutput>>;
```
jmoseley, line 143: Is it possible to _not_ do this? is it necessary for `wrapFunction` to _always_ return a promise, regardless of the function it is wrapping?

jmoseley (commented):
In my current in-flight change to capture stream results, I am moving to a model where `Component` does not require the underlying function to be async (it can wrap either or) and does not return a promise in the non-promise case. We don't have any code in the tracing stuff that needs to be async. The only async things are initializing the context and waiting for all updates to flush, and we can do both of those things inside the `Workflow` wrapper, so `Component` remains non-async.

jmoseley:
That may need to change for durability, but I would rather introduce it if needed, as opposed to changing the signature of any wrapped SDKs.

examples/deep-research/src/firecrawlProvider.ts
dereklegenzoff: good call!

packages/gensx-core/src/wrap.ts
dereklegenzoff: Yeah I think I can remove this and some of the other changes. The reason this got added was to 1) fix linting and 2) make sure stuff returned the right type (i saw some scenarios where my other code changes led to us returning `promise<string | promise<string>>` and this was how i attempted to fix that. 

The only wrap change i absolutely need is this so that i can wrap methods that don't have any inputs:
```
export function wrapFunction<TInput extends object = object, TOutput = unknown>(
  fn: ((input: TInput) => Promise<TOutput> | TOutput) | ((input?: TInput) => Promise<TOutput> | TOutput),
```

The only implication is that linting will be broken when wrapping a sync function with no inputs but if you're planning to make additional changes seems like i shouldn't worry about the linting for now.

packages/gensx-core/src/component.ts
```diff
@@ -63,46 +63,46 @@ function getResolvedOpts(
...
     }
 
-    return createWorkflow(target, decoratorOpts);
+    return createWorkflow<P, R>(target, decoratorOpts);
   };
 }
 
-export function createComponent<P extends object, R>(
+export function createComponent<P extends object = {}, R = unknown>(
   target: (props: P) => R,
```
jmoseley, line 106: I think this needs to be optional to support functions that take no parameters.

packages/gensx-core/tests/component.test.ts
```diff
@@ -75,6 +75,40 @@ suite("component", () => {
     // In the real implementation, it will use the function name
   });
 
+  test("can handle components with no arguments (undefined props)", async () => {
```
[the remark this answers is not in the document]
dereklegenzoff: good call, added.

packages/gensx-core/src/component.ts
jmoseley: Very weird. Ok seems fine then.

PR #707, opened 2025-05-27
feat: removing decorators

dereklegenzoff:


packages/gensx-core/src/component.ts
```diff
@@ -62,62 +61,11 @@ function getResolvedOpts(
...
-    }
-
-    return createWorkflow<P, R>(target, decoratorOpts);
-  };
-}
-
-export function createComponent<P extends object = {}, R = unknown>(
+export function Component<P extends object = {}, R = unknown>(
+  name: string,
   target: (props: P) => R,
```
jmoseley, line 66: Is it possible to widen this type? Accept any function with any shape of args?

I guess we would need to change the tracing API...

PR #721, opened 2025-05-29
feat: updating ai rules to programming model v2

dereklegenzoff:


packages/gensx-claude-md/templates/CLAUDE.md
```diff
@@ -36,84 +34,123 @@ npm run start
...
-    </OpenAIProvider>
-  )
-);
+const greeting = await Greeting({ name: "John", formal: true });
 ```
 
-### Component Nesting with Child Functions
+### Workflow Definition
+
+In GenSX, workflows are the entry points to your application. Here's a basic example of a GenSX workflow:
```
dereklegenzoff, line 79: I don't love this verbiage but couldn't really think of anything better at the moment; would love suggestions.

EvanBoyle: Here is a suggestion on how to differentiate the two: 

- Components are reusable workflow steps. They are a unit of tracing, and the inputs and outputs of the component will be recorded and traced
- Workflows are collections of components. Each workflow will create a top level trace that shows all components that executed within it. Workflows are automatically converted into API endpoints for synchronous and asyncronous invocation when exported from the file (show example).

PR #756, opened 2025-06-09
feat: streaming structured data & react integration

EvanBoyle:
The goal of this PR is to improve GenSX react integration by having a story for defining state in the server, and then streaming that state over the wire, and consuming it via a well typed react hook. 

This PR adds: 

- gensx primitives for state: `workflowState` `componentState` `StatefulComponent` (and tests!)
- `@gensx/react` package for the new react hooks
- blog-writer-react example that shows the hole experience e2e
- various updates and bug fixes to the dev server to make it work well inside of a next.js project

I still have a few TODOs I'm cleaning up, but it is ready enough to have folks start to take a look. 

One thing we should discuss is the name for "State", and if we want to call it something else like "Status". 

The `state-design.md` file is the document that I used with cursor as I was designing this. I will leave it for now so that folks can read it for an overview, but I will remove it before merging.


EvanBoyle:
Closing this out for now. Will layer some pieces back in in separate PRs, but we decided not to go with hierarchical state and adding new concepts like StatefulComponent

PR #757, opened 2025-06-10
Add public option to project config

jmoseley:
## Summary
- allow project visibility via `public` flag in config and CLI
- parse project config with js-yaml
- support `--public` option in `new` command
- test project creation with `--public` flag

## Testing
- `pnpm format`
- `pnpm lint`
- `pnpm test` *(fails: create-gensx tests)*

------
https://chatgpt.com/codex/tasks/task_e_6841b7e7efb0832c9d301e10e557f7ab


packages/gensx-core/src/utils/project-config.ts
```diff
@@ -1,13 +1,15 @@
...
 
+import { load } from "js-yaml";
 import { z } from "zod";
 
 // Define schema for gensx.yaml
 const ProjectConfigSchema = z.object({
   projectName: z.string(),
   environmentName: z.string().optional(),
   description: z.string().optional(),
+  public: z.boolean().optional(),
```
dereklegenzoff, line 12: seems like there's a mismatch between here and the schema we use in the gensx package? `public` vs `publicWorkflows`

PR #759, opened 2025-06-11
fix: Improve local dev experience

jmoseley:
## Proposed changes

* Fixes to support building when inside a next project
* Improvements for deploying from inside a next project (also enable yarn/pnpm support)
* Fix progress events for the local dev server

### DIFF
--- a/containers/builder/build.sh
+++ b/containers/builder/build.sh
@@ -18,10 +18,16 @@ done
 
 cd /tmp/project
 
-# Install required dependencies
-npm install
+# Use yarn or pnpm if there is a yarn.lock or pnpm-lock.yaml
+if [ -f "yarn.lock" ]; then
+  yarn install
+elif [ -f "pnpm-lock.yaml" ]; then
+  pnpm install
+else
+  npm install
+fi
 
-WORKFLOW_PATH=${WORKFLOW_PATH:-"workflows.tsx"}
+WORKFLOW_PATH=${WORKFLOW_PATH:-"workflows.ts"}
 
 # Build with ncc - this should bundle all dependencies
 ncc build ./${WORKFLOW_PATH} -o /out/dist --target es2022 -e "@libsql/client"
--- a/packages/gensx/src/commands/start.tsx
+++ b/packages/gensx/src/commands/start.tsx
@@ -75,14 +75,28 @@ export const StartUI: React.FC<Props> = ({ file, options }) => {
       throw new Error("Could not find tsconfig.json");
     }
 
-    let tsconfig;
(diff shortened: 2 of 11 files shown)


containers/builder/build.sh
```diff
@@ -18,10 +18,16 @@ done
...
 cd /tmp/project
 
-# Install required dependencies
-npm install
-
-WORKFLOW_PATH=${WORKFLOW_PATH:-"workflows.tsx"}
+# Use yarn or pnpm if there is a yarn.lock or pnpm-lock.yaml
+if [ -f "yarn.lock" ]; then
+  yarn install
+elif [ -f "pnpm-lock.yaml" ]; then
```
EvanBoyle, line 24: nit: should we default to preferred tools first in case both exist (ie use pnpm by default)

packages/gensx/src/commands/start.tsx
```diff
@@ -75,14 +75,28 @@ export const StartUI: React.FC<Props> = ({ file, options }) => {
...
+      tsconfig.options.incremental = false;
+      tsconfig.options.noEmit = false;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      tsconfig.raw.compilerOptions.incremental = false;
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      tsconfig.raw.compilerOptions.noEmit = false;
+      tsconfig.options.outDir = ".gensx/dist";
+      tsconfig.options.target = ts.ScriptTarget.ES2015;
+      tsconfig.options.module = ts.ModuleKind.NodeNext;
+      tsconfig.options.moduleResolution = ts.ModuleResolutionKind.NodeNext;
```
EvanBoyle, line 99: Do you think this will lead to weird breaks or edge cases? 

For instance, does this require `.js` imports for files that are related to the gensx deployment, but potentially not for others in the project?

EvanBoyle (approved):

packages/gensx/src/commands/start.tsx
jmoseley: Had an idea out on my bike ride. I'll just add type: module to the package.json, and see how that works. We have the commonjs compatibility examples to validate against.

EvanBoyle: I did add `type: module` to my nextjs package.json. It didn't break the next build / deployment, but I'm not 100% clear on the downstream implications of doing so on behalf of the user.

jmoseley: Oh we shouldn't make it a permanent change, just for the build. My understanding is that it just affects the output (changing it from commonjs to ESM).

PR #761, opened 2025-06-12
feat: Added sdk, custom hook, and example app (draft pad)

danfhernandez:


packages/gensx-client/example.ts
```diff
@@ -0,0 +1,226 @@
+/**
```
EvanBoyle, line 1: Should probably remove this file?

packages/gensx-client/src/sdk.ts
```diff
@@ -0,0 +1,601 @@
...
+    this.org = config.org;
+    this.project = config.project;
+    this.environment = config.environment;
+  }
+
+  /**
+   * Run a workflow with optional streaming
+   * @returns Either { output, progressStream } or { outputStream, progressStream } based on stream flag
+   */
+  async run<TOutput = unknown>(
```
EvanBoyle, line 207: run should be generic over <TInput, TOutput>, no? 

Then RuntimeOptions should also be generic over the input type. `RuntimeOptions<TInput>`

packages/gensx-react/src/hooks/use-gensx.ts
```diff
@@ -0,0 +1,498 @@
...
+ *   project: 'different-project',
+ *   inputs: { userMessage: 'Hello from different org' }
+ * });
+ * ```
+ */
+export function useWorkflow<
+  TInputs = any,
+  TOutput = any,
+  // This should actually be BaseProgressEvent
+  TEvent extends GenSXProgressEvent = GenSXProgressEvent,
```
EvanBoyle, line 171: What do we need this generic for? If it isn't essential, might be best to remove.

packages/gensx-react/src/hooks/use-gensx.ts
```diff
@@ -0,0 +1,498 @@
...
+  progressEvents: TEvent[];
+
+  /** Output events */
+  outputEvents: GenSXOutputEvent[];
+
+  /** Run workflow in collection mode (returns final output) */
+  run: (inputs: TInputs) => Promise<TOutput | null>;
+
+  /** Run workflow in streaming mode */
+  stream: (inputs: TInputs) => Promise<void>;
```
EvanBoyle, line 121: @jmoseley I'm curious if it feels right to you to expose separate methods for stream and run? 

Side note, is `run()` actually `/start`? or is it actually `/run`?

packages/gensx-react/src/hooks/use-gensx.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  /** Stop the current workflow */
+  stop: () => void;
```
EvanBoyle, line 124: Does this work? Do we have a /cancel API to call?

EvanBoyle (commented):
First round of notes and questions. Looking good so far!

packages/gensx-react/src/hooks/use-gensx.ts
danfhernandez: I just removed stream and made it so it always "streams" which really isn't any different than run for the user

EvanBoyle (commented):
We need to update the `release-please-config.json` so that we'll publish to npm on merge.

packages/gensx-react/index.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+// Export hooks
+export * from './src/hooks/use-gensx';
+
+// Example placeholder exports
+export const Button = ({ children, ...props }: React.ButtonHTMLAttributes<HTMLButtonElement>) => {
+  return <button {...props}>{children}</button>;
+};
```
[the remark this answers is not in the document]
EvanBoyle: @danfhernandez why do we need to keep the button in here?

packages/gensx-client/CHANGELOG.md
jmoseley, line 1: Lets get rid of this file, let release-please create it.

packages/gensx-client/src/sdk.ts
```diff
@@ -0,0 +1,370 @@
...
+
+  /**
+   * Run a workflow and return the raw Response object
+   * Provides direct access to the fetch response without any processing
+   *
+   * @param workflowName - Name of the workflow to run
+   * @param options - Options including format: 'sse' | 'ndjson' | 'json'
+   * @returns Raw Response object
+   */
+  async runRaw(
```
jmoseley, line 187: Would it make sense to have an easy-to-use `run` for immediately getting the progress? Or perhaps returning a listener we can bind to?

```
const execution = await gensxSdk.run(...);
execution.on('progress', (event: GenSXEvent) => {
   ...
});
```

We should consider that API for `getProgress` too.

packages/gensx-react/package.json
```diff
@@ -0,0 +1,64 @@
...
+  "dependencies": {
+    "@gensx/client": "workspace:*"
+  },
+  "devDependencies": {
+    "@types/react": "^19",
+    "@types/react-dom": "^19",
+    "typescript": "^5"
+  },
+  "peerDependencies": {
+    "react": "^19.0.0"
```
jmoseley, line 59: Do we really need to be on React 19 for this?

packages/gensx-react/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "lint:fix": "eslint . --fix",
+    "type-check": "tsc --noEmit",
+    "clean": "rm -rf dist"
+  },
+  "keywords": [
+    "gensx",
+    "react",
+    "hooks",
+    "ai",
+    "jsx"
```
jmoseley, line 46:
```suggestion
```

packages/gensx-client/.npmrc
```diff
@@ -0,0 +1,2 @@
+access=public
```
jmoseley, line 1: I don't think this file is necessary

packages/gensx-react/.npmrc
```diff
@@ -0,0 +1,2 @@
+access=public
```
jmoseley, line 1: I don't think this file is necessary

packages/gensx-react/src/hooks/use-gensx.ts
```diff
@@ -0,0 +1,437 @@
...
+export type JsonValue =
+  | string
+  | number
+  | boolean
+  | null
+  | JsonValue[]
+  | { [key: string]: JsonValue };
+
+// WorkflowMessage type (from gensx-core)
+export type WorkflowMessage =
```
jmoseley, line 20: Can't this be imported from @gensx/client?

jmoseley (approved):
A few things, LGTM o/w.

Also update `update-example-deps.sh` with the new package names.

packages/gensx-react/package.json
danfhernandez: Moved down to 18

packages/gensx-client/src/sdk.ts
danfhernandez: I think for this version I'm tempted to push as is and then we can follow up with that run capability for the sdk and layer it into the hook

danfhernandez: That would be really nice

examples/draft-pad/app/api/gensx/route.ts
```diff
@@ -0,0 +1,156 @@
...
+  return true;
+};
+
+/**
+ * API route that acts as a pure passthrough to GenSX
+ * Accepts the same parameters as the GenSX SDK
+ *
+ * This is designed to work with the useGenSX hook
+ */
+export async function POST(request: NextRequest) {
```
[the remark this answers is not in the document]
danfhernandez: That would be really slick. Perhaps we do this in a follow up. Captured here: https://github.com/gensx-inc/gensx/issues/789

PR #764, opened 2025-06-12
feat: Add helpers for event stream and state.

jmoseley:
## Proposed changes

* change `emitProgress` to `publishData`
* add `publishEvent` to publish arbitrary events with a given label.
* add `publishObject` to publish arbitrary objects with a given label.
* add `createEventStream` as a helper for creating a strongly typed event emitter
* add `createObjectStream` as a helper for creating a strongly typed state emitter

Things to call out:
* I opted not to use JSON patch or some sort of updating mechanism for transmitting state changes. That would make the stream a stateful thing (must process the whole thing when reconnecting), and I don't want to necessarily make that required yet.
* `emitProgress` becomes a bit of a second class citizen with its new name, and I think that is the right place for it. It is fully-flexible, but not really the happy path that we will document front and center.
* Because we are not doing anything smart about updating objects partially, Objects and Events are really the same, just with different names. How we process those things (the React hooks) is where they will differ.

### DIFF
--- a/packages/gensx-core/src/workflow-state.ts
+++ b/packages/gensx-core/src/workflow-state.ts
@@ -1,0 +1,79 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-type-parameters */
+import { getCurrentContext } from "./context.js";
+
+// JSON-serializable value type for progress data
+export type JsonValue =
+  | string
+  | number
+  | boolean
+  | null
+  | JsonValue[]
+  | { [key: string]: JsonValue };
+
+export type WorkflowMessage =
+  | { type: "start"; workflowExecutionId?: string; workflowName: string }
+  | {
+      type: "component-start";
+      componentName: string;
+      label?: string;
+      componentId: string;
+    }
+  | {
+      type: "component-end";
+      componentName: string;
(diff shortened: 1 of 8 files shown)


packages/gensx-core/src/workflow-state.ts
```diff
@@ -0,0 +1,107 @@
...
+  };
+}
+
+/**
+ * Create a function that publishes a state to the workflow message stream with the given label.
+ *
+ * @param label - The label of the state.
+ * @returns A function that publishes a state to the workflow message stream.
+ */
+export function createWorkflowState<
```
dereklegenzoff, line 101: I'm not in love with the names `createWorkflowState` and `publishState`. The name `createWorkflowState` makes me think I'm creating state that i'm going to be using within my workflow.  I'd personally go for `createObjectStream` and `publishObject`. Not super strongly opinionated here though.

dereklegenzoff (approved):

packages/gensx-core/src/workflow-state.ts
jmoseley: I'm coming around to this, I agree with you.

PR #783, opened 2025-06-14
feat: enable checkpoint replay

EvanBoyle:
WIP branch for enabling checkpoint replay. A checkpoint JSON can be passed into a workflow as a `workflowOpt` and be used as a "cache" to skip over components that have already executed. 

Updated checkpoint construction to insert "cached" nodes into the new checkpoint if they are found during replay.

Updated component ID generation to be deterministic to support this. Added tests for it all. Some TODOs remaining: 

1. ID generation needs more improvement. I expect we'll discover bugs when we start writing tests against things like JS builtins (`new Date()`), and we will probably need to have stronger and more deterministic props hashing vs just using `JSON.stringify`. 
3. I will add some tools like `gensx.createCheckpoint()` that can be used for testing and debugging (write a checkpoint out to a file so that it is easier to debug, etc).


packages/gensx-core/src/checkpoint.ts
```diff
@@ -1,61 +1,36 @@
...
+  parentId?: string,
+): string {
+  // Simple but effective serialization for MVP
+  const propsStr = JSON.stringify(props, Object.keys(props).sort());
+  const propsHash = createHash("sha256")
+    .update(propsStr)
+    .digest("hex")
+    .slice(0, 16);
+
+  return `${parentId ?? "root"}:${name}:${propsHash}`;
```
jmoseley, line 33: One thing this does not take into account (and I did not solve in my prototype either) is multiple child components with the props. We basically make the assumption that they are all the identical in that case.

I think we could solve this via a sequence counter, that increase for each component, either at the parent or workflow level, and include that in the hash (or it is the whole hash?).

packages/gensx-core/src/component.ts
```diff
@@ -88,6 +90,26 @@ export function Component<P extends object = {}, R = unknown>(
       );
     }
 
+    // Generate deterministic ID for replay
+    const props_for_id = props
+      ? Object.fromEntries(
+          Object.entries(props).filter(
+            ([key]) => key !== "children" && key !== "componentOpts",
```
jmoseley, line 97: We shouldn't see these in the props anymore, right?

packages/gensx-core/src/checkpoint.ts
jmoseley: Those could work. A sequence number would be make it no longer a concern of the user. Javascript iteration is deterministic, so it should be reliable unless it is non-deterministic code.

EvanBoyle: Are you sure we can guarantee sequence in the case where components are created in a loop and awaited via promise.all?

packages/gensx-core/src/checkpoint.ts
```diff
@@ -26,11 +26,11 @@ export function generateDeterministicId(
   // Simple but effective serialization for MVP
   const propsStr = JSON.stringify(props, Object.keys(props).sort());
   const propsHash = createHash("sha256")
-    .update(propsStr)
+    .update(`${propsStr}:${parentId ?? "root"}`)
     .digest("hex")
     .slice(0, 16);
 
-  return `${parentId ?? "root"}:${name}:${propsHash}`;
+  return `${name}:${propsHash}`;
```
jmoseley, line 33: If we include `parentId` in the ID, IDs will get arbitrarily long, relative to the nesting.

EvanBoyle: This is true.

The impact of including 
1. "Identity" is stronger - fewer surprises when components that match props but not parent get skipped during checkpoint replay
2. checkpoints get (exponentially) bigger the more you nest
3. Not practical to store these IDs in a database unless we hash them -- not sure if we need to. 

Are you comfortable making this simplification? Without framework limits, we'll probably still have problems storing the name-props-only version of this ID in the database without hashing.

packages/gensx-core/src/component.ts
```diff
@@ -207,6 +207,7 @@ export function Component<P extends object = {}, R = unknown>(
     }
 
     try {
+      // TODO: Don't emit this when rerunning the workflow with a partial checkpoint.
```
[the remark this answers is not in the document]
EvanBoyle: I think at a minimum I just need to understand more about the mechanics of having a redis stream active for a long period of time.. ie hours / days / weeks. And going between the states of redis => s3 => redis again, and whether or not this process will be reliable.

packages/gensx-core/src/wait-for-input.ts
```diff
@@ -0,0 +1,40 @@
...
+
+// TODO: Is there a security issue here? Does this endpoint need to be authenticated?
+function getCallbackUrl() {
+  return `${process.env.GENSX_API_BASE_URL}/org/${process.env.GENSX_ORG}/workflowExecutions/${process.env.GENSX_EXECUTION_ID}/runs/${process.env.GENSX_EXECUTION_RUN_ID}/resume`;
+}
+
+// TODO: Name this better
+export async function waitForInput<T extends Record<string, unknown>>(
+  trigger: (callbackUrl: string) => Promise<void>,
+  // schema: z.ZodSchema<T>, // TODO
```
EvanBoyle, line 12: 👍

packages/gensx-core/src/wait-for-input.ts
```diff
@@ -0,0 +1,40 @@
...
+): Promise<T> {
+  // TODO: We should do some locking here to prevent multiple simultaneous waitForInput calls.
```
EvanBoyle, line 14: Locking is good, but even more so we need to design a shutdown routine:

1. Do we let outstanding component trees resolve to completion? 
2. Do we kill outstanding components? 
3. Do we take a final checkpoint and reject any updates that come in after the fact? 

Answers to these questions are very tightly coupled to the guarantees we are able to provide (certain choices could move us from exactly once to at most once...)

EvanBoyle (commented):
@jmoseley tagging you here since I assume you aren't subscribed to notifications (I opened the original PR). 

Left a handful of comments but overall things are looking good so far.

jmoseley (approved):

PR #786, opened 2025-06-16
feat: adding project cli commands

### LINKED ISSUE #782: Project management in the CLI
We should be able to create projects directly from the CLI.

For example, if I manually create a gensx.yaml file I am unable to create the project without deploying.

dereklegenzoff:
Adds CLI commands for projects:
- `gensx project` - shows the current project. includes details on existing envs and workflows
- `gensx project create` - create a project
- `gensx project ls` - lists projects for the org

https://github.com/user-attachments/assets/5c852c7b-3473-47af-a961-74f85d91a012

Fixes #782

### DIFF
--- a/packages/gensx/src/commands/project/create.tsx
+++ b/packages/gensx/src/commands/project/create.tsx
@@ -1,0 +1,397 @@
+import { Box, Text, useApp } from "ink";
+import SelectInput from "ink-select-input";
+import TextInput from "ink-text-input";
+import { useEffect, useState } from "react";
+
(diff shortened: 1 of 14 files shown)


jmoseley (approved):
Whats the behavior if you are creating a project in a folder that already has a gensx.yaml?

dereklegenzoff:
> Whats the behavior if you are creating a project in a folder that already has a gensx.yaml?

For both `gensx project create` and `gensx project` we by default will try to use the project name/description in gensx.yaml but you can override override it via `gensx project create myProject` and `gensx project -p myProject`

PR #791, opened 2025-06-18
docs: updated draft pad example

danfhernandez:
Updated the draft pad example to be a more robust example. It now allows you to run up to 9 different models in parallel for a given request. It tracks price, speed, words, etc for each and fully streams each output independently. 

See demo here: https://screen.studio/share/ZJRXb9eg

Data is pull using the new models.dev endpoint.


packages/gensx-core/src/workflow-state.ts
```diff
@@ -38,13 +38,13 @@ export interface WorkflowDataMessage {
...
   type: "event";
-  data: Record<string, JsonValue>;
+  data: JsonValue;
   label: string;
 }
 
 export interface WorkflowObjectMessage {
   type: "object";
-  data: Record<string, JsonValue>;
+  data: JsonValue;
```
jmoseley, line 47: Are we ok with a valid value being `string` or `number`, etc?

dereklegenzoff: I think that's fine. I could see a scenario where the object you're streaming is the status of the workflow and it's just a string or things like that. Really any type you could use with `useState` feels reasonable.

jmoseley: Sounds good. These are written in at least three places, ill look at making a types package.

examples/draft-pad/DEPLOYMENT.md
```diff
@@ -0,0 +1,187 @@
...
+
+Since the GenSX CLI needs authentication in the CI environment, create a `.gensx/credentials.json` file structure:
+
+```bash
+mkdir -p .gensx
+```
+
+Then create a build script that sets up credentials:
+
+The `scripts/setup-gensx-ci.sh` script will automatically set up GenSX credentials from environment variables during the build process.
```
dereklegenzoff, line 64: this script doesn't exist.

jmoseley, line 64: What is this?

examples/draft-pad/DEPLOYMENT.md
```diff
@@ -0,0 +1,187 @@
...
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Deploy GenSX Workflow
+        env:
+          GENSX_TOKEN: ${{ secrets.GENSX_TOKEN }}
+          GENSX_API_KEY: ${{ secrets.GENSX_API_KEY }}
+        run: |
+          chmod +x ./scripts/setup-gensx-ci.sh
```
jmoseley, line 96: This script does not exist.

We should build an action.

examples/draft-pad/app/api/models/route.ts
```diff
@@ -0,0 +1,56 @@
...
+function isProviderAvailable(provider: string): boolean {
+  const requiredEnvVars = PROVIDER_ENV_REQUIREMENTS[provider] ?? [];
+  return requiredEnvVars.every((envVar) => !!process.env[envVar]);
+}
+
+export async function GET() {
+  try {
+    const response = await fetch("https://models.dev/api.json", {
+      // Add cache control to avoid hitting the API too frequently
+      next: { revalidate: 3600 }, // Cache for 1 hour
```
jmoseley, line 27: I'm not finding any documentation about this `next` option for fetch.

jmoseley: Oh here it is: https://nextjs.org/docs/app/api-reference/functions/fetch#optionsnextrevalidate

Very cool.

examples/draft-pad/DEPLOYMENT.md
danfhernandez: I removed this file. Has a lot of errant crap in there.

examples/draft-pad/scripts/deploy.sh
```diff
@@ -0,0 +1,45 @@
...
+set -e
+
+# Check for required environment variables
+if [ -z "$GENSX_API_KEY" ]; then
+  echo "❌ Error: GENSX_API_KEY environment variable is required"
+  exit 1
+fi
+
+# Build the command with required variables
+CMD="npx gensx@latest deploy ./gensx/workflows.ts --project draft-pad --env production --yes"
```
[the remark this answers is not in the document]
danfhernandez: Updated.

packages/gensx-core/src/workflow-state.ts
```diff
@@ -123,9 +119,7 @@ export function publishObject<
  * @param label - The label of the event.
  * @returns A function that publishes an event to the workflow message stream.
  */
-export function createEventStream<
-  T extends Record<string, JsonValue> = Record<string, JsonValue>,
->(label: string) {
+export function createEventStream<T = JsonValue>(label: string) {
```
jmoseley, line 122: ditto


```suggestion
export function createEventStream<T extends JsonValue = JsonValue>(label: string) {
```

jmoseley (approved):

PR #793, opened 2025-06-19
feat: Checkpoint restoration

jmoseley:
## Proposed changes

Support restoring a prior checkpoint and restarting workflow execution from that point, including feedback from "the future".

```ts
export const ComponentWithCheckpoint = gensx.Component("ComponentWithCheckpoint", async ({ input }: Props) => {
  const { feedback, restore } = gensx.createCheckpoint();
  gensx.publishData(`Got feedback: ${feedback}`);

  // Wait for feedback from the user
  const result = await gensx.waitForInput<{ data: string }>(
    async (callbackUrl) => {
      gensx.publishData(callbackUrl);
      await setTimeout(1000);
    },
  );
  gensx.publishData(`Got result from external system: '${result.data}'`);

  // If the user does not approve, restore the checkpoint and include the feedback from the user.
  if (result.data.includes("error")) {
    gensx.publishData("Restoring from checkpoint");
    await restore(result.data);
  }

  return result.data;
});
```

### DIFF
--- a/packages/gensx-core/src/restore-checkpoint.ts
+++ b/packages/gensx-core/src/restore-checkpoint.ts
@@ -22,8 +22,6 @@ const CheckpointMarkerComponent = Component(
   },
 );
 
-const CheckpointMarkerLabels = new Map<string, string>();
-
 export function createCheckpoint<T = unknown>(
   {
     label,
@@ -39,7 +37,9 @@ export function createCheckpoint<T = unknown>(
   label: string;
 } {
   label ??= `checkpoint-marker-${Date.now()}`;
-  if (CheckpointMarkerLabels.has(label)) {
+  const context = getCurrentContext();
+  const workflowContext = context.getWorkflowContext();
+  if (workflowContext.checkpointLabelMap.has(label)) {
     throw new Error(`[GenSX] Checkpoint ${label} has already been created.`);
   }
   // Do not pass the label as a prop, as that would affect that ability to deterministically calculate the nodeId.
@@ -49,7 +49,7 @@ export function createCheckpoint<T = unknown>(
       metadata: { label, maxRestores },
     },
   );
-  CheckpointMarkerLabels.set(label, result.nodeId);
+  workflowContext.checkpointLabelMap.set(label, result.nodeId);
 
   if (result.restoreCount >= result.maxRestores) {
     throw new Error(
@@ -84,9 +84,14 @@ export async function restoreCheckpoint<T = unknown>(
   feedback: T,
 ) {
   // TODO: Add some locking mechanism to prevent multiple simultaneous restores, or for other workflow work to happen while we're restoring.
-  if (!CheckpointMarkerLabels.has(label)) {
+  const context = getCurrentContext();
+  const workflowContext = context.getWorkflowContext();
+  if (!workflowContext.checkpointLabelMap.has(label)) {
     throw new Error(`[GenSX] Checkpoint ${label} has not been created.`);
   }
 
-  await restoreCheckpointByNodeId(CheckpointMarkerLabels.get(label)!, feedback);
+  await restoreCheckpointByNodeId(
+    workflowContext.checkpointLabelMap.get(label)!,
+    feedback,
+  );
 }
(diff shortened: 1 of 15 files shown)


packages/gensx-core/src/wait-for-input.ts
```diff
@@ -2,8 +2,8 @@ import { Component } from "./component.js";
 import { getCurrentContext } from "./context.js";
 
 // TODO: Is there a security issue here? Does this endpoint need to be authenticated?
-function getCallbackUrl() {
-  return `${process.env.GENSX_API_BASE_URL}/org/${process.env.GENSX_ORG}/workflowExecutions/${process.env.GENSX_EXECUTION_ID}/runs/${process.env.GENSX_EXECUTION_RUN_ID}/resume`;
+function getCallbackUrl(nodeId: string) {
+  return `${process.env.GENSX_API_BASE_URL}/org/${process.env.GENSX_ORG}/workflowExecutions/${process.env.GENSX_EXECUTION_ID}/resume/${nodeId}`;
 }
 
 // TODO: Name this better
```
EvanBoyle, line 9: Should we go ahead and rename this to `pauseForInput`?

packages/gensx-core/src/workflow-context.ts
```diff
@@ -30,6 +34,15 @@ export function createWorkflowContext({
       (async () => {
         console.warn("[GenSX] Pause/resume not supported in this environment");
       }),
+    onRestoreCheckpoint:
+      onRestoreCheckpoint ??
+      // eslint-disable-next-line @typescript-eslint/require-await
+      (async () => {
+        console.warn(
```
EvanBoyle, line 41: Should open an issue to update the dev server to support pause + checkpoint restore.

EvanBoyle: But I'm OK with holding off on that for now if we think we'll need to iterate on the APIs a bit.

packages/gensx-core/tests/restore-checkpoint.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      expect(consoleErrorSpy).toHaveBeenCalledWith(
+        "[GenSX] Restoring checkpoints is not supported in this environment.",
+      );
+    });
+
+    consoleErrorSpy.mockRestore();
+  });
+
+  test("createCheckpoint respects maxRestores option", () => {
```
EvanBoyle, line 164: Feels like we should have a test that ensures that a workflow can create multiple checkpoints and restore from each of them.

packages/gensx-core/src/wait-for-input.ts
```diff
@@ -12,18 +12,23 @@ export async function waitForInput<T extends Record<string, unknown>>(
   // schema: z.ZodSchema<T>, // TODO
 ): Promise<T> {
   // TODO: We should do some locking here to prevent multiple simultaneous waitForInput calls.
+  const TriggerComponent = Component(
+    "WaitForInputTrigger",
+    async ({ nodeId }: { nodeId: string }) => {
```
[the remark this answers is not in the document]
EvanBoyle: Ah, I see that we are including the parentID in the hash at least.

packages/gensx-core/src/restore-checkpoint.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  if (result.restoreCount >= result.maxRestores) {
+    throw new Error(
+      `[GenSX] Checkpoint ${label} has been restored more than ${result.maxRestores} times.`,
+    );
+  }
+
+  return {
+    feedback: result.feedback as T | null,
+    restore: (feedback: T) =>
```
EvanBoyle, line 62: Is it safe to pass this `restore` function around in the workflow? IE can I pass it to another component? 

If the program crashes and replays will this function still be "intact"? IE `json.parse` on the outputs of this component actually rehydrates the restore function...

packages/gensx-core/src/wait-for-input.ts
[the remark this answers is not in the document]
jmoseley: And we have sequence numbers in the nodeId as well.

packages/gensx-core/src/workflow-context.ts
jmoseley: https://github.com/gensx-inc/gensx/issues/796

PR #807, opened 2025-06-21
fix: Always build with type:module.

jmoseley:
## Proposed changes

The bundler we use (ncc) seems to have trouble properly resolving ESM modules if `"type": "module"` is not set in the package.json, as is common for NextJS projects. This change manually sets it during build time to try and work around it.


dereklegenzoff (approved):

jmoseley:
It would not surprise me if we see weird build errors in some number of commonjs projects due to this.

PR #822, opened 2025-06-25
Improve API error messages in storage

dereklegenzoff:
## Summary
- parse API error messages from response JSON
- show detailed errors for blob, database, and search
- update unit tests for new error handling

## Testing
- `npx turbo run test --filter=./packages/gensx-storage`

------
https://chatgpt.com/codex/tasks/task_e_685c35627c6083259f1d747c4f22bccc

### DIFF
--- a/packages/gensx-storage/src/utils/parse-error.ts
+++ b/packages/gensx-storage/src/utils/parse-error.ts
@@ -1,0 +1,11 @@
+export async function parseErrorResponse(response: Response): Promise<string> {
+  try {
+    const data = (await response.json()) as { error?: string };
+    if (data && typeof data.error === "string" && data.error.trim()) {
+      return data.error;
+    }
+  } catch {
+    // Ignore JSON parse errors
+  }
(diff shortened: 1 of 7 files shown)


packages/gensx-storage/src/utils/parse-error.ts
```diff
@@ -0,0 +1,11 @@
+export async function parseErrorResponse(response: Response): Promise<string> {
+  try {
+    const data = (await response.json()) as { error?: string };
```
jmoseley, line 3: If the caller eventually tries to call `response.json()` on this, it will error because the body was already read. Safer to not mutate the input and clone the response.

```suggestion
  try {
    const clone = response.clone();
    const data = (await clone.json()) as { error?: string };
```

jmoseley (approved):

packages/gensx-storage/src/utils/parse-error.ts
dereklegenzoff: good call, updated

PR #830, opened 2025-06-30
feat: Support front end tool calling

jmoseley:
## Proposed changes

* The `external-tool` message, that triggers a front end tool call (via `@gensx/react`)
* The `Toolbox` type, to share tools between the front- and back-end
* A demo app to demonstrate how to use and implement front end tools
* chore: Refactor the dev server
* fix: Extract the sequence number from node IDs to better support replays on changing source code

This is not quite ready yet. The demo is not working quite yet because I haven't fully flushed out checkpoint replay with streaming nodes. That is a work in progress.


packages/gensx-vercel-ai/package.json
```diff
@@ -52,15 +52,17 @@
...
     "@vitest/coverage-istanbul": "catalog:",
     "vitest": "catalog:",
-    "zod-to-json-schema": "catalog:"
+    "zod": "catalog:"
   },
   "publishConfig": {
     "access": "public"
   },
+  "peerDependencies": {
+    "zod": ">=3.25.56"
```
dereklegenzoff, line 61: nice! Hopefully this prevents all the nasty zod version issues

packages/gensx-vercel-ai/src/tools.ts
```diff
@@ -0,0 +1,29 @@
...
+    (acc, [name, toolDef]) => {
+      let paramsSchema;
+      if ("_zod" in toolDef.params) {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any
+        paramsSchema = jsonSchema(z4.toJSONSchema(toolDef.params) as any);
+      } else {
+        paramsSchema = toolDef.params;
+      }
+      acc[name] = {
+        description: toolDef.description ?? `Tool: ${name}`,
```
dereklegenzoff, line 19: nit: description is optional ([docs](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#tools.tool.description)) so it's probably best not to muck with it 
```suggestion
        description: toolDef.description,
```

dereklegenzoff: ahh i see, you're doing this because it's optional in the AI SDK but not on our end. I think we should probably just make it optional on our end to align with what the ai sdk does.

examples/fe-tools/gensx/agent.ts
```diff
@@ -0,0 +1,236 @@
...
+      ],
+    });
+
+    // Add system message for geographic focus
+    const systemMessage: CoreMessage = {
+      role: "system",
+      content: `You are a helpful geographic assistant that can interact with an interactive map. You have access to several map tools:
+
+- webSearch: Search the web for information relevant to the user's query
+- geocode: Geocode a location from an address or a query to a specific location, returned with latitude and longitude, as well as other useful information about the location
```
dereklegenzoff, line 187: I'd probably define this in `workflows.ts` instead so we can easily rip out this code once we have the agent abstraction.

In `loadChatHistory()` you can add the system message to the messages if history is undefined: `return history ?? [systemMessage];`

examples/fe-tools/demo.md
```diff
@@ -0,0 +1,130 @@
+# Interactive Map Chat Demo
```
dereklegenzoff, line 1: maybe combine this with the readme?

examples/fe-tools/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@examples/fe-tools",
```
dereklegenzoff, line 2: nit: i'd probably rename this example to `front-end-tools` or `client-side-tools` just to make it stand out a bit more.

packages/gensx-react/src/hooks/use-gensx.ts
```diff
@@ -200,10 +227,62 @@ export function useWorkflow<TInputs = unknown, TOutput = unknown>(
             setInProgress(false);
             onError?.(event.error);
             break;
+
+          case "external-tool":
```
dereklegenzoff, line 231: will we use this case to handle human in the loop or do we need a separate case for that?

packages/gensx-vercel-ai/src/tools.ts
```diff
@@ -0,0 +1,29 @@
+import type { InferToolParams, ToolBox } from "@gensx/core";
+
+import { executeExternalTool } from "@gensx/core";
+import { jsonSchema, type Tool } from "ai";
+import z4 from "zod/v4";
+
+// Utility function to convert ToolBox to Vercel AI SDK ToolSet format
+export function asToolSet(toolBox: ToolBox): Record<string, Tool> {
```
dereklegenzoff, line 8: I suppose we'll need similar functions for the openai and anthropic packages eventually?

dereklegenzoff (approved):
Overall this all looks really good and some really nice refactoring included as well.

I wonder if we should standardize on calling these `client-side tools` (sounds a bit cleaner than front-end tools and aligns with the verbiage we've seen used)

packages/gensx-react/src/hooks/use-gensx.ts
jmoseley: Human in the loop is upto the caller. They provide a `trigger` function which gets the callback URL as an argument, and inside that function they  push that callback URL to whatever system they need.

EvanBoyle (approved):

PR #841, opened 2025-07-09
fix: Avoid duplication issue due to object mutation.

jmoseley:
## Proposed changes

Ensure that we appropriately clone objects before mutating them.

### DIFF
--- a/packages/gensx-react/src/hooks/use-object.ts
+++ b/packages/gensx-react/src/hooks/use-object.ts
@@ -1,0 +1,172 @@
+import type {
+  JsonValue,
+  Operation,
+  WorkflowMessage,
+  WorkflowObjectMessage,
+} from "@gensx/core";
+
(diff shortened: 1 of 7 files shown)


packages/gensx-react/src/hooks/use-object.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/**
+ * Apply a JSON patch to reconstruct object state. This is useful for consumers who want to reconstruct the full object state from patches.
+ *
+ * @param patches - The JSON patch operations to apply.
+ * @param currentState - The current state of the object (defaults to empty object).
+ * @returns The new state after applying the patches.
+ */
+export function applyObjectPatches(
```
dereklegenzoff, line 78: do we actually want to export this?

dereklegenzoff (approved):

EvanBoyle (approved):

PR #859, opened 2025-07-16
feat: adding next.js template to `gensx new`

dereklegenzoff:
Adds in next.js template option to gensx new

Template selector:
Screenshot 2025-07-16 at 11 07 29 AM

New ouput:
Screenshot 2025-07-16 at 11 07 15 AM

Basic template app:
template-ux

Light mode:
[
Screenshot 2025-07-16 at 12 10 04 PM
](url)


jmoseley (approved):
I think for the next big template we should look at something like sao, hygen, or even degit as a way to make the templates cleaner.

dereklegenzoff:
> I think for the next big template we should look at something like sao, hygen, or even degit as a way to make the templates cleaner.

That sounds like a good idea--i was able to avoid renaming everything ".template" until i got to getting the gensx-cli package working. perhaps there was some setting i could have enabled but gave up after a few attempts.

PR #870, opened 2025-07-17
fix: Fix issues with zod/firecrawl deps.

jmoseley:
## Proposed changes

* Firecrawl-js > 1.21 requires Node 22 for some readon
* I got ahead of ourselves trying to have good support for Zod 4. Backed it off a bit more to keep it compatible with the AI SDK.

### DIFF
--- a/packages/gensx-core/package.json
+++ b/packages/gensx-core/package.json
@@ -50,7 +50,7 @@
     "@common.js/serialize-error": "^11.0.3",
     "deterministic-object-hash": "^2.0.2",
     "ini": "^5.0.0",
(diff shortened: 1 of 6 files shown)


packages/gensx-core/package.json
```diff
@@ -50,7 +50,7 @@
     "@common.js/serialize-error": "^11.0.3",
     "deterministic-object-hash": "^2.0.2",
     "ini": "^5.0.0",
-    "zod": "^3.25.0 || ^4.0.0",
```
jmoseley, line 53: If we use 4, we break support for the AI sdk.

PR #881, opened 2025-07-18
feat: Add action for building examples on PR.

jmoseley:
## Proposed changes

- Add action for building example workflows on PR
- Fix path filters to match the right thing.

### DIFF
--- a/.github/workflows/build-example-workflows.yml
+++ b/.github/workflows/build-example-workflows.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
     paths:
       - "examples/client-side-tools/gensx/**"
-      - ".github/workflows/build-example-workflows.yml"
 
 jobs:
   build-client-side-tools:
--- a/.github/workflows/build-example-workflows.yml
+++ b/.github/workflows/build-example-workflows.yml
@@ -1,0 +1,23 @@
+name: Build GenSX Example Workflows
+
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-example-workflows.yml
```diff
@@ -1,10 +1,10 @@
-name: Deploy GenSX Examples
+name: Deploy GenSX Example Workflows
 
 on:
   push:
     branches: [main]
     paths:
-      - "examples/client-side-tools/gensx"
+      - "examples/client-side-tools/gensx/**"
```
dereklegenzoff, line 7: funny enough - i had this exact same issue with my yml file. Must be a weak point of claude.

.github/workflows/build-example-workflows.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "examples/client-side-tools/gensx/**"
+
+jobs:
+  build-client-side-tools:
+    runs-on: ubuntu-latest
+    steps:
```
dereklegenzoff, line 11: Should we do something like this so we can reduce redunancy? Only tricky bit is that web apps vs basic examples have a slightly different file path `src/workflow.ts` vs `gensx/workflow.ts`

```
    strategy:
      matrix:
        path: [deep-research, client-side-tools]
```

PR #884, opened 2025-07-18
docs: Add docs for humand-in-the-loop and client side tool calling.

jmoseley:
## Proposed changes

Add docs for human-in-the-loop and client-side tool calling.


website/docs/src/content/client-side-tools.mdx
```diff
@@ -0,0 +1,396 @@
+---
+title: Client-Side Tools
+description: Execute functions in the browser from your workflows
+---
+
+# Client-Side Tools
+
+Client-side tools let your workflows execute functions directly in the browser or client application. This is useful for interacting with client-side APIs, updating UI state, or accessing browser features like geolocation.
```
EvanBoyle, line 8: How about something like: 

Client-side tools are for cases where a tool can’t run on the same server that’s calling the LLM API. Instead, the tool needs to execute in the user’s environment—like a browser, a desktop app, or any other remote client. Use this when the function needs to touch client-only APIs (e.g. geolocation), update UI state, search through code on a users local machine, or interact with something the server can’t access directly.

website/docs/src/content/durable-execution.mdx
```diff
@@ -0,0 +1,326 @@
+---
+title: Durable Execution
+description: Build resilient workflows with pause/resume, retries, and human-in-the-loop capabilities
+---
+
+# Durable Execution
+
+GenSX provides durable execution capabilities that make your workflows resilient and allow for complex orchestration patterns. This includes state preservation, checkpoint restoration, and seamless human-in-the-loop integration.
```
EvanBoyle, line 8: Build workflows that don’t break. GenSX gives you first-class primitives for pause/resume, retries, and human input—without losing state or writing scaffolding code.

Durable execution is built in. State is preserved automatically, workflows resume after failure, and you can checkpoint anywhere in the flow to rewind with new input or retry logic.

website/docs/src/content/durable-execution.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+## Key capabilities
```
EvanBoyle, line 10: I'd just remove this heading or change it to "Overview".

website/docs/src/content/durable-execution.mdx
```diff
@@ -0,0 +1,326 @@
...
+
+### State preservation
+
+All workflow state is automatically preserved across restarts and failures:
```
EvanBoyle, line 14: GenSX automatically snapshots all workflow state between component steps. If the process crashes or restarts, it picks up right where it left off:

website/docs/src/content/durable-execution.mdx
```diff
@@ -0,0 +1,326 @@
...
+    }
+    
+    return "User data processing cancelled";
+  }
+);
+```
+
+## Human-in-the-loop integration
+
+Durable execution seamlessly integrates with human-in-the-loop patterns. See the [Human-in-the-Loop](/human-in-the-loop) documentation for detailed examples.
```
EvanBoyle, line 45: Need to wait for someone to click a button, review a doc, or approve a change? GenSX makes it easy to pause execution until human input arrives—with zero polling or timers. See the [Human-in-the-Loop](/human-in-the-loop) documentation for detailed examples.

EvanBoyle (commented):
Started reviewing files individually but decided that the tone was a bit off for the durable and human in the loop docs and decided to try and do a small rewrite: 

https://gist.github.com/EvanBoyle/57c9853e964b8ae0afbd86cca6cd4519

EvanBoyle (approved):
🚢 

One of the docs still has a "key capabilities" header without any subtext which IMO feels awkward (we could just remote it) but other than that looks good!

PR #900, opened 2025-07-22
feat: Support executionScope.

jmoseley:
## Proposed changes

Add support for execution scope, that can be provided by an api key that is bound to a specific execution scope (such as a `userId` or `threadId`).


dereklegenzoff (approved):
A bit hard to grok what this would look like in practice but overall lgtm.

jmoseley:
> A bit hard to grok what this would look like in practice but overall lgtm.

Yeah I bet we will use it a bit and tweak the API. I'm going to hold off on documenting it until after I integrate it into one or two examples.

PR #907, opened 2025-07-28
fix: make @libsql/client optional in @gensx/storage

### LINKED ISSUE #788: @gensx/storage can't be used in next apps
The @libsql/client package we leverage for filesystem dbs causes errors in next.js

The current workaround is to update the next.config.ts to this:

```
/** @type {import('next').NextConfig} */
const nextConfig = {
  eslint: {
    ignoreDuringBuilds: true,
  },
  images: { unoptimized: true },
  webpack: (config) => {
    // Ignore @libsql/client package
    config.resolve.alias = {
      ...config.resolve.alias,
      "@libsql/client": false,
    };
    return config;
  },
};

module.exports = nextConfig;
```

dereklegenzoff:
## Summary
- make libsql an optional peer dependency
- load libsql dynamically for filesystem databases
- update docs for optional installation and Next.js config

fixes #788

### DIFF
--- a/packages/gensx-storage/README.md
+++ b/packages/gensx-storage/README.md
@@ -8,35 +8,11 @@
 npm install @gensx/storage
 ```
 
-If you plan to use local databases when developing, install `@libsql/client` as well:
+If you plan to use local databases when developing, install `@libsql/client` as a dev dependency:
 
 ```bash
 npm install -d @libsql/client
 ```
-
-## Next.js Configuration
-
-When using local databases in a Next.js project you should exclude `@libsql/client` from client bundles. Add the following to your `next.config.ts` or `next.config.js` file:
-
-```typescript
-/** @type {import('next').NextConfig} */
-const nextConfig = {
-  // ... other config options
-  webpack: (config: any) => {
-    // Ignore @libsql/client package for client-side builds
-    config.resolve.alias = {
-      ...config.resolve.alias,
-      "@libsql/client": false,
-    };
-    return config;
-  },
-  // ... other config options
-};
-
-module.exports = nextConfig;
-```
-
-This configuration prevents bundling issues when `@libsql/client` is installed. If you only use cloud databases you can skip installing `@libsql/client` and omit this webpack rule. See the [client-side-tools example](https://github.com/gensx-inc/gensx/tree/main/examples/client-side-tools) for a complete implementation.
 
 ## Features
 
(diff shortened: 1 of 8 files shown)


packages/gensx-storage/package.json
```diff
@@ -58,7 +58,12 @@
   },
   "dependencies": {
     "@gensx/core": "workspace:*",
-    "@turbopuffer/turbopuffer": "^0.10.4",
+    "@turbopuffer/turbopuffer": "^0.10.4"
+  },
+  "peerDependencies": {
+    "@libsql/client": "^0.15.9"
+  },
```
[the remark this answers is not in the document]
dereklegenzoff: added the devDependency. stuck with peer dependencies and made it optional since true optional dependencies are auto-installed

packages/gensx-storage/src/database/databaseClient.ts
```diff
@@ -15,30 +13,57 @@ import {
...
+      return this.storage;
+    }
+
+    if (this.storagePromise) {
+      return this.storagePromise;
+    }
+
+    this.storagePromise = this.initializeStorage();
+    this.storage = await this.storagePromise;
+    return this.storage;
```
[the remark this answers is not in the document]
dereklegenzoff: fixed

packages/gensx-storage/README.md
```diff
@@ -245,3 +227,29 @@ const results = await namespace.query({
   topK: 3,
 });
 ```
+
+## Troubleshooting
+
+If you have `@libsql/client` installed as a dependency, it may cause issues for bundlers. You should always install it as a dev dependency as local databases can only be used during development.
```
jmoseley, line 233: > can only be used during development

Is that actually true? Aren't there cases where someone might want to use a filesystem database (eg production, running on their own server or something)?

packages/gensx-storage/README.md
```diff
@@ -245,3 +227,29 @@ const results = await namespace.query({
...
+    };
+    return config;
+  },
+  // ... other config options
+};
+
+module.exports = nextConfig;
+```
+
+This configuration prevents bundling issues when `@libsql/client` is installed. If you only use cloud databases you can skip installing `@libsql/client` and omit this webpack rule. See the [client-side-tools example](https://github.com/gensx-inc/gensx/tree/main/examples/client-side-tools) for a complete implementation.
```
jmoseley, line 255: I think this example will be out of date once this is merged, right?

jmoseley (approved):

packages/gensx-storage/README.md
dereklegenzoff: true - once we publish this update i was planning to go simplify the config in each of the files.

PR #910, opened 2025-07-31
feat: quickstart for templates and gensx example commands

dereklegenzoff:
- Adds CLI commands for examples:
   - `gensx examples` lists available examples
   - `gensx examples clone <example-nam>` clones an example
   - Adds docs for new CLI commands
- Adds quickstart on using templates
   - Include tabs to make the page customized by exaple


packages/gensx/src/commands/examples/clone.tsx
```diff
@@ -0,0 +1,131 @@
...
+        if (fs.existsSync(targetPath)) {
+          throw new Error(
+            `Directory ${targetDir} already exists. Please choose a different name or remove the existing directory.`,
+          );
+        }
+
+        setStep("Cloning example...");
+
+        // Use degit to clone the example
+        const degitCommand = `npx degit ${example.path} ${targetDir}`;
```
jmoseley, line 60: Does this assume they have cloned the gensx repo? The `path` in the supported examples collection is just a relative path, so how does this work for `npx gensx examples clone draft-pad`?

jmoseley: NVM I misunderstood how degit works.

PR #927, opened 2025-08-13
feat: Mastra example

jmoseley:
## Proposed changes

Proof of concept to run mastra on our infra. Next steps would be to move the wrapper into a package, and add better instrumentation utilities/integrations.

Changes to packages:
* Support plain callables as exports. Since the mastra workflows are sort of dynamic and not statically analyzable we need to support this to generate the schemas
* Make workflow name an explicit property (`__gensxWorkflowName`). This is bundler safe if we ever change the bundling method in the future, and a consistent property we can look at in both the dev server and the deployed workflows.
* Support env variables for project and environment for configuring checkpointing (related to https://github.com/gensx-inc/gensx/issues/912)

### DIFF
--- a/examples/mastra/.gitignore
+++ b/examples/mastra/.gitignore
@@ -1,0 +1,8 @@
+output.txt
+node_modules
+dist
+.mastra
+.env.development
+.env
+*.db
+*.db-*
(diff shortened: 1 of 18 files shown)


dereklegenzoff (approved):
Pretty neat. Example is missing a readme but otherwise looks good.

jmoseley:
hello from genie

PR #928, opened 2025-08-13
chore: Support node 24 and deprecate node 18.

jmoseley:
## Proposed changes

Remove testing for node 18, and add testing for node 24. At some point we should consider bumping min node version for packages too.


jmoseley:
hello from genie

